PR #60, opened 2021-03-09
Remove experimental gradle tasks from doc

rsinukov:
These tasks were broken for some time and it doesn't worth keeping them up to date on our side. Gradle config for samples was reworked here https://github.com/ktorio/ktor-samples/pull/89 and tasks are now obsolete.

### DIFF
--- a/topics/http-client_multiplatform.md
+++ b/topics/http-client_multiplatform.md
@@ -78,24 +78,4 @@ dependencies {
 
 There is a full sample using the common client in the ktor-samples repository [mpp/client-mpp](https://github.com/ktorio/ktor-samples/tree/1.3.0/mpp/client-mpp).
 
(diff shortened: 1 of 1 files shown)


topics/http-client_multiplatform.md
```diff
@@ -78,24 +78,4 @@ dependencies {
 
 There is a full sample using the common client in the ktor-samples repository [mpp/client-mpp](https://github.com/ktorio/ktor-samples/tree/1.3.0/mpp/client-mpp).
```
andreyaksenov, line 79: Shouldn't we change `1.3.0` to `master` after merging https://github.com/ktorio/ktor-samples/pull/89 ?

andreyaksenov (changes requested):

PR #61, opened 2021-03-10
Client overview

andreyaksenov:
Wrote a new 'Client Overview' topic, the result is here: https://helpserver.labs.jb.gg/help/ktor/1.5.2/client.html
Also created a small example demonstrating the `Logging` feature for using it in this new topic and the old `Logging` topic: [client-logging](https://github.com/ktorio/ktor-documentation/tree/client-overview/codeSnippets/snippets/client-logging).


topics/client.md
```diff
@@ -1,105 +1,204 @@
...
-[`OkHttp`](http-client_engines.md#okhttp),
-[`Android`](http-client_engines.md#android),
-[`Ios`](http-client_engines.md#ios),
-[`Js`](http-client_engines.md#js-javascript),
-[`Jetty`](http-client_engines.md#jetty),
-[`CIO`](http-client_engines.md#cio) and [`Mock`](http-client_engines.md#mock). For a full list of dependencies required for a specific engine, see [](http-client_engines.md#dependencies). 
 
-For example, you can add a `CIO` engine dependency as follows:
+### Engine Dependency {id="engine-dependency"}
+An engine is in charge of processing network requests. There are different client engines available for [various platforms](http-client_multiplatform.md), such as Apache, CIO, Android, iOS, and so on. For example, you can add a `CIO` engine dependency as follows:
```
hhariri, line 16: in charge -> responsible.

topics/client.md
```diff
@@ -1,105 +1,204 @@
...
+In this example, we use the [CIO](https://api.ktor.io/%ktor_version%/io.ktor.client.engine.cio/-c-i-o/index.html) engine. You can also omit an engine:
 
 ```kotlin
+import io.ktor.client.*
+
 val client = HttpClient()
 ```
 
-Ktor will choose an engine among the ones that are available from the included artifacts using a `ServiceLoader` on the JVM, or similar approach in the other platforms. If there are multiple engines in the dependencies Ktor chooses first in alphabetical order of engine name.
+In this case, the client will choose an engine automatically depending on the artifacts [added in a build script](#engine-dependency). You can learn more from the [Engines](http-client_engines.md#default) topic.
```
hhariri, line 47: "You can learn more from the Engines topic" doesn't sound quite right.

topics/client.md
```diff
@@ -1,105 +1,204 @@
...
 
-It's safe to create multiple instance of client or use the same client for multiple requests.
+## Configure the Client {id="configure-client"}
 
-## Releasing resources
+### Basic Configuration {id="basic-config"}
 
-Ktor client is holding resources: prepared threads, coroutines and connections. After you finish working with the client, you may wish to release it by calling `close`:
+To configure the client, you can pass an additional functional parameter to the client constructor. The [HttpClientEngineConfig](https://api.ktor.io/%ktor_version%/io.ktor.client.engine/-http-client-engine-config/index.html) class is a base class for configuring the client. In the example below, receiving HTTP errors in response don't cause exceptions:
```
hhariri, line 53: This seems a bit strange. "In the exmaple below,..."

maybe indicate "For instance, we can change the behaviour of exceptions in the Basic Configuration:"

topics/client.md
```diff
@@ -1,105 +1,204 @@
...
+import io.ktor.client.engine.cio.*
+
+val client = HttpClient(CIO) {
+    expectSuccess = false
+}
 ```
 
-If you want to use a client to make only one request consider `use`-ing it. The client will be automatically closed once the passed block has been executed:
+### Engine Configuration {id="engine-config"}
+You can configure an engine using the `engine` method in a block:
```
hhariri, line 65: "`engine" method in a block" sounds strange. What do you mean by block?

topics/client.md
```diff
@@ -1,105 +1,204 @@
...
+    }
 }
 ```
 
-The method `close` signals to stop executing new requests. It wouldn't block and allows all current requests to finish successfully and release resources. You can also wait for closing with the `join` method or halt any activity using the `cancel` method. For example:
+See [Engines](http-client_engines.md) section for additional details.
+
+### Features {id="features"}
 
+Ktor lets you use additional client functionality (features) that is not available by default, for example, logging, authorization, or serialization. Most of them are distributed in separate artifacts. For example, you can log HTTP calls by installing the [Logging](features_logging.md) feature:
```
hhariri, line 82: distributed -> provided

topics/client.md
```diff
@@ -1,105 +1,204 @@
...
+You can also configure a feature inside the `install` block. For example, for the `Logging` feature, you can specify the logger and logging level:
+```kotlin
+```
+{src="snippets/client-logging/src/main/kotlin/com/example/Application.kt" include-symbol="client"}
 
-Ktor HttpClient follows `CoroutineScope` lifecycle. Check out [Coroutines guide](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-scope/) to learn more.
+## Make a Request {id="make-request"}
 
-## Client configuration
+The main way for making HTTP requests is the [request](https://api.ktor.io/%ktor_version%/io.ktor.client.request/request.html) function that takes URL as a parameter. Inside this function, you can configure various request parameters: specify an HTTP method, add headers, specify the request body, and so on. These parameters are exposed by the [HttpRequestBuilder](https://api.ktor.io/%ktor_version%/io.ktor.client.request/-http-request-builder/index.html) class.
```
hhariri, line 99: takes a URL

topics/client.md
```diff
@@ -1,105 +1,204 @@
 [//]: # (title: Client Overview)
```
hhariri, line 1: All samples are using code directly. They should be using code snippet references.

topics/client.md
```diff
@@ -1,105 +1,204 @@
...
+    body = TextContent(
+        text = "Body content",
+        contentType = ContentType.Text.Plain
+    )
+}
+```
+
+
+### Receive a Response {id="response"}
+Functions used to make a request (`request`, `get`, `post`, etc.) allow you to receive a response in several ways:
```
hhariri, line 169: All request functions allow ....

topics/features_logging.md
```diff
@@ -17,20 +17,15 @@ To enable logging, you need to include the following artifacts in the build scri
```
hhariri, line 17: Should be using code snippet

hhariri (changes requested):

topics/client.md
andreyaksenov: > What do you mean by block?

I mean 'the `engine` method block'. Does it sound better? Or better to simply write 'pass the engine configuration to the `engine` method'?

topics/client.md
andreyaksenov: what's wrong here? I need to clarify what a reader can learn?

topics/client.md
andreyaksenov: done

topics/client.md
hhariri: I find it the construct somewhat strange. What are they meant to be learning more on? What engines are or how they work? 

If it's about what engines are, then I'd suggest something "If you want to learn more about engines, please check the [Engine topic]

although I'd replace topic for documentation TBH.

topics/client.md
hhariri: I think we should name that what they are which is functions.

hhariri (changes requested):
Updated.

topics/client.md
andreyaksenov: ok, I see. Changed to 'You can learn how the client chooses an engine from the *** documentation section.'

hhariri (approved):

PR #70, opened 2021-03-18
Distribution with the Gradle Application plugin

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-2305
This is a draft version for preliminary review.

### DIFF
--- a/codeSnippets/snippets/gradle-app/initial.gradle
+++ b/codeSnippets/snippets/gradle-app/initial.gradle
@@ -1,0 +1,14 @@
+plugins {
+    id 'org.jetbrains.kotlin.jvm' version '%kotlin_version%'
+}
+
+group 'org.example'
+version '1.0-SNAPSHOT'
+
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    implementation "org.jetbrains.kotlin:kotlin-stdlib"
+}
--- a/codeSnippets/snippets/gradle-app/kotlin/app.kt
+++ b/codeSnippets/snippets/gradle-app/kotlin/app.kt
@@ -1,0 +1,18 @@
+// src/main/kotlin/app.kt
+
+import io.ktor.application.*
+import io.ktor.response.*
+import io.ktor.routing.*
+import io.ktor.server.cio.*
+import io.ktor.server.engine.*
+
+fun main() {
+    val server = embeddedServer(CIO, port = 8080) {
+        routing {
+            get("/") {
+                call.respondText { "Hello" }
+            }
+        }
+    }
+    server.start(wait = true)
+}
--- a/ktor.tree
+++ b/ktor.tree
@@ -162,6 +162,8 @@
         <toc-element toc-title="Deployments"
                      id="deploy.md">
             <toc-element toc-title="Packing">
+                <toc-element id="application-gradle-plugin.md"
+                             toc-title="Gradle Application plugin"/>
                 <toc-element id="fatjar.md"
                              accepts-web-file-names="packing.html"/>
                 <toc-element id="war.md" toc-title="WAR"/>
--- a/topics/application-gradle-plugin.md
+++ b/topics/application-gradle-plugin.md
@@ -1,0 +1,68 @@
+[//]: # (title: Distribution via Gradle Application plugin)
+
+From the Gradle [documentation](https://docs.gradle.org/current/userguide/application_plugin.html):
+> The Application plugin facilitates creating an executable JVM application. 
+> It makes it easy to start the application locally during development, 
+> and to package the application as a TAR and/or ZIP including operating system specific start scripts.
+
+We will use the second responsibility of this plugin and show how to package a simple Ktor application.
+
+Let's start with the following project structure:
+
+```text
+├── build.gradle
+├── settings.gradle
+└── src
+    └── main
+        ├── kotlin
+        └── resources
+├── gradle
+├── gradlew
+└── gradlew.bat
+```
+You can use the [wrapper](https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:adding_wrapper) task to generate `gradle`, `gradlew` and `gradlew.bat` files.
+
+Here is the initial content for the build configuration:
+
+<tabs>
+    <tab title="Gradle (Groovy)">
+        <code style="block" lang="Groovy" interpolate-variables="true" src="snippets/gradle-app/initial.gradle" />
+    </tab>
+</tabs>
+
+To use Ktor we need to add corresponding dependencies:
+
+<tabs>
+    <tab title="Gradle (Groovy)">
+        <code style="block" lang="Groovy" interpolate-variables="true" src="snippets/gradle-app/ktor-deps.gradle" />
+    </tab>
+</tabs>
+
+Then create a file with the simplest possible code for our server application:
+<code style="block" lang="Kotlin" src="snippets/gradle-app/kotlin/app.kt" />
+
+Now we need to run our application to make sure it works. We can do it by clicking on a [gutter icon](https://www.jetbrains.com/help/idea/settings-gutter-icons.html) in Intellij IDEA or 
+by building project and running it with `java`.
+
+The next step is to actually apply the Gradle Application plugin and set the main class for it:
+
+<tabs>
+    <tab title="Gradle (Groovy)">
+        <code style="block" lang="Groovy" src="snippets/gradle-app/plugin.gradle" />
+    </tab>
+</tabs>
+
+Finally, run `distZip` or `distTar` task to package our application:
+
+<tabs>
+    <tab title="Linux/MacOS">
+        <code style="block" lang="Bash">./gradlew distZip</code>
+    </tab>
+    <tab title="Windows">
+        <code style="block" lang="CMD">gradlew.bat distZip</code>
+    </tab>
+</tabs>
+
+By default, the plugin places the resulting archive in the `build/distributions` directory. 
+To run the application after archive distribution, extract its contents and 
+execute the start script from the `bin` directory for your OS.
--- a/codeSnippets/snippets/gradle-app/ktor-deps.gradle
+++ b/codeSnippets/snippets/gradle-app/ktor-deps.gradle
@@ -1,0 +1,8 @@
+ext.ktor_version = "%ktor_version%"
+
+dependencies {
+    // ...
+    implementation "io.ktor:ktor-server-core:$ktor_version"
+    implementation "io.ktor:ktor-server-cio:$ktor_version"
+    implementation "ch.qos.logback:logback-classic:1.2.3"
+}
--- a/codeSnippets/snippets/gradle-app/plugin.gradle
+++ b/codeSnippets/snippets/gradle-app/plugin.gradle
@@ -1,0 +1,8 @@
+plugins {
+    // ...
+    id 'application'
+}
+
+application {
+    mainClass = 'AppKt'
+}


codeSnippets/snippets/gradle-app/initial.gradle
```diff
@@ -0,0 +1,14 @@
+plugins {
```
hhariri, line 1: Is the reason for splitting the Gradle file into multiple so as to be able to reference them in the docs by sections?  If so, I wonder how this impacts the usability of a person that wants to just use this sample by clicking on the button to download it. It might feel overwhelmingly complex.

codeSnippets/snippets/gradle-app/kotlin/app.kt
```diff
@@ -0,0 +1,18 @@
+// src/main/kotlin/app.kt
+
+import io.ktor.application.*
+import io.ktor.response.*
+import io.ktor.routing.*
+import io.ktor.server.cio.*
+import io.ktor.server.engine.*
+
+fun main() {
+    val server = embeddedServer(CIO, port = 8080) {
```
hhariri, line 10: Not sure if we should be using CIO, especially since we don't recommend it yet for production.  Also, we should probably avoid the embeddedServer and use HOCON like most other snippets.

ktor.tree
```diff
@@ -162,6 +162,8 @@
         <toc-element toc-title="Deployments"
```
hhariri, line 162: Is the plan to do a separate one for Maven? If not, then maybe file should be named something more generic. Also, I think the title should reflect what the intention is as opposed to the name of the plugin, i.e. "Distributing Applications using Gradle" 

And lastly, I think the top-level TOC we should probably also rename to Application Distribution

topics/application-gradle-plugin.md
```diff
@@ -0,0 +1,68 @@
+[//]: # (title: Distribution via Gradle Application plugin)
+
+From the Gradle [documentation](https://docs.gradle.org/current/userguide/application_plugin.html):
+> The Application plugin facilitates creating an executable JVM application. 
+> It makes it easy to start the application locally during development, 
+> and to package the application as a TAR and/or ZIP including operating system specific start scripts.
+
+We will use the second responsibility of this plugin and show how to package a simple Ktor application.
```
hhariri, line 8: not sure we need to explicitly say "second responsibility"

topics/application-gradle-plugin.md
```diff
@@ -0,0 +1,68 @@
...
+├── settings.gradle
+└── src
+    └── main
+        ├── kotlin
+        └── resources
+├── gradle
+├── gradlew
+└── gradlew.bat
+```
+You can use the [wrapper](https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:adding_wrapper) task to generate `gradle`, `gradlew` and `gradlew.bat` files.
```
hhariri, line 23: We <> You. The paragraph above using We, in this one You. We need to keep it consistent. @andreyaksenov have we settled on We or You?

topics/application-gradle-plugin.md
```diff
@@ -0,0 +1,68 @@
...
+> The Application plugin facilitates creating an executable JVM application. 
+> It makes it easy to start the application locally during development, 
+> and to package the application as a TAR and/or ZIP including operating system specific start scripts.
+
+We will use the second responsibility of this plugin and show how to package a simple Ktor application.
+
+Let's start with the following project structure:
+
+```text
+├── build.gradle
```
hhariri, line 13: Does this reflect the actual distribution that is being used? Cause I see three Gradle files.

topics/application-gradle-plugin.md
```diff
@@ -0,0 +1,68 @@
...
+    └── main
+        ├── kotlin
+        └── resources
+├── gradle
+├── gradlew
+└── gradlew.bat
+```
+You can use the [wrapper](https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:adding_wrapper) task to generate `gradle`, `gradlew` and `gradlew.bat` files.
+
+Here is the initial content for the build configuration:
```
hhariri, line 25: Here -> Below

topics/application-gradle-plugin.md
```diff
@@ -0,0 +1,68 @@
...
+
+<tabs>
+    <tab title="Gradle (Groovy)">
+        <code style="block" lang="Groovy" interpolate-variables="true" src="snippets/gradle-app/initial.gradle" />
+    </tab>
+</tabs>
+
+To use Ktor we need to add corresponding dependencies:
```
hhariri, line 33: add -> add the

topics/application-gradle-plugin.md
```diff
@@ -0,0 +1,68 @@
...
+
+<tabs>
+    <tab title="Gradle (Groovy)">
+        <code style="block" lang="Groovy" interpolate-variables="true" src="snippets/gradle-app/ktor-deps.gradle" />
+    </tab>
+</tabs>
+
+Then create a file with the simplest possible code for our server application:
```
hhariri, line 41: Then create a file .... 
-> 

Four our sample we're using a simple Ktor server that responds to a URL with some text:

topics/application-gradle-plugin.md
```diff
@@ -0,0 +1,68 @@
...
+<code style="block" lang="Kotlin" src="snippets/gradle-app/kotlin/app.kt" />
+
+Now we need to run our application to make sure it works. We can do it by clicking on a [gutter icon](https://www.jetbrains.com/help/idea/settings-gutter-icons.html) in Intellij IDEA or 
+by building project and running it with `java`.
+
+The next step is to actually apply the Gradle Application plugin and set the main class for it:
```
hhariri, line 47: I'd better phrase it as:

Gradle Application plugin requires us to set the main class for the application. In our case it would be: 


This clearly indicates that the plugin requires this.

topics/application-gradle-plugin.md
```diff
@@ -0,0 +1,68 @@
...
+<tabs>
+    <tab title="Linux/MacOS">
+        <code style="block" lang="Bash">./gradlew distZip</code>
+    </tab>
+    <tab title="Windows">
+        <code style="block" lang="CMD">gradlew.bat distZip</code>
+    </tab>
+</tabs>
+
+By default, the plugin places the resulting archive in the `build/distributions` directory.
```
hhariri, line 66: Maybe worth mention that you can also use the installDist to avoid having to unzip.

hhariri (changes requested):

codeSnippets/snippets/gradle-app/initial.gradle
Stexxe: The reason is having steps that are easier to understand. I can refer to the whole Gradle build file after.

codeSnippets/snippets/gradle-app/kotlin/app.kt
Stexxe: The goal of this article is to show how to package an application and the choice of an engine isn't relevant here. Again, I see no difference between the usage of `embeddedServer` and `HOCON` here, so I chose the simplest one.

ktor.tree
Stexxe: > Is the plan to do a separate one for Maven?

Yes, in the separate article.

topics/application-gradle-plugin.md
Stexxe: It's the initial project structure to start with

codeSnippets/snippets/gradle-app/initial.gradle
hhariri: Right, so the project is going to have multiple Gradle build files then?

Stexxe: Yes

codeSnippets/snippets/gradle-app/kotlin/app.kt
hhariri: I mention this for two reasons:

1. I think it would make sense to have consistency across our code snippets
2. In real world applications, people need to deploy other files, and resources is a good example. In fact I'd go as far as saying your sample should include examples of using `static` too so that people understand where these files are copied and distributed to.

Stexxe: 1. We have two up-to-date methods. Why in the documentation should I prefer only one? 
2. I agree about showing packing resources but that's not about `HOCON`

hhariri: Not about preference, but about consistency. In any case no objections, but would be good to show then examples with static.

topics/application-gradle-plugin.md
```diff
@@ -0,0 +1,68 @@
+[//]: # (title: Distribution via Gradle Application plugin)
+
+From the Gradle [documentation](https://docs.gradle.org/current/userguide/application_plugin.html):
```
andreyaksenov, line 3: I'd not cite Gradle docs because they can be changed. Also, tips are intended to add some additional info or hints and don't suit for citation.
It's better to add a clear introduction and describe which this help topic is aimed for. It's OK to copy/paste from their docs and write smth like this: `The Gradle [Application plugin] (https://docs.gradle.org/current/userguide/application_plugin.html) facilitates ... . In this topic, we'll show how to ... .`

topics/application-gradle-plugin.md
andreyaksenov: +1

topics/application-gradle-plugin.md
andreyaksenov: > Let's start with the following project structure

What does this mean? I should generate an empty Gradle project by myself? Why `snippets/gradle-app` is not a runnable project and doesn't reflect this structure?

topics/application-gradle-plugin.md
andreyaksenov: I use `You`

topics/application-gradle-plugin.md
andreyaksenov: And comma after `To use Ktor`. 
We->You

topics/application-gradle-plugin.md
```diff
@@ -0,0 +1,68 @@
...
+<tabs>
+    <tab title="Gradle (Groovy)">
+        <code style="block" lang="Groovy" interpolate-variables="true" src="snippets/gradle-app/ktor-deps.gradle" />
+    </tab>
+</tabs>
+
+Then create a file with the simplest possible code for our server application:
+<code style="block" lang="Kotlin" src="snippets/gradle-app/kotlin/app.kt" />
+
+Now we need to run our application to make sure it works. We can do it by clicking on a [gutter icon](https://www.jetbrains.com/help/idea/settings-gutter-icons.html) in Intellij IDEA or
```
andreyaksenov, line 44: > in Intellij IDEA

We are talking about the Idea for the first time here. I think a reader can get lost. If we organize this topic as a step-by-step tutorial, it's better to have some `Prerequisites` (IDEA installed, a link to the initial project, etc.) and then describe the actual steps (1,2,3) to allow a reader follow us.

topics/application-gradle-plugin.md
andreyaksenov: +1
We can also add a link to a new [Docker](https://ktor.io/docs/docker.html#building-and-running-the-docker-image) topic since it use `installDist`.

andreyaksenov (changes requested):

codeSnippets/snippets/gradle-app/kotlin/app.kt
andreyaksenov: I also vote for consistency:
1. This example should be runnable.
2. Have the same structure as other examples from the `snippets` folder.
3. Have readme. 
You can still keep the initial Gradle file and include Ktor dependencies using the `lines` attribute (it works now).

andreyaksenov:
Closing, the resulting topic is here: https://ktor.io/docs/gradle-application-plugin.html

PR #79, opened 2021-04-02
Fix kotlinx html

andreyaksenov:
- changed ktor version to 1.5.3
- added links to space packages to fix the issue with 'kotlinx-html'

### DIFF
--- a/codeSnippets/gradle.properties
+++ b/codeSnippets/gradle.properties
@@ -5,9 +5,9 @@ kotlin.code.style = official
 org.gradle.configureondemand = false
 
 # versions
-kotlin_version = 1.4.30
-ktor_version = 1.5.2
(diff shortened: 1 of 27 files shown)


codeSnippets/gradle.properties
```diff
@@ -5,9 +5,9 @@ kotlin.code.style = official
 org.gradle.configureondemand = false
 
 # versions
-kotlin_version = 1.4.30
-ktor_version = 1.5.2
-kotlinx_coroutines_version = 1.3.9-native-mt-2
```
e5l, line 10: native-mt

e5l (approved):

codeSnippets/gradle.properties
andreyaksenov: thanks, fixed

PR #80, opened 2021-04-04
Updating WebSocket testing methods

aspanu:
Some browsers seem to not allow messages to be sent from https://www.websocket.org/echo.html to `localhost` using `ws`. They only allow `wss` connections. I added a small note to clarify that for future readers in case they get stuck on it like I did.


aspanu:
Note: I'm not affiliated in any way with the Chrome extension I suggested in my PR, but it has been working well for me so far.

PR #87, opened 2021-04-14
Client request

andreyaksenov:
Rewrote the 'Sending requests' topic for a client and added two runnable code samples used in this topic:
* [client-upload-file](https://github.com/ktorio/ktor-documentation/tree/client-request/codeSnippets/snippets/client-upload-file)
* [client-parallel-requests](https://github.com/ktorio/ktor-documentation/tree/client-request/codeSnippets/snippets/client-parallel-requests)


hhariri:
hey, a topic for this PR can be found here:
https://helpserver.labs.jb.gg/help/ktor/1.5.3/request.html
Yaroslav said it's better not to reveal our infrastructure details in
public repos


Regards,
Hadi Hariri
http://hadihariri.com | @hhariri

topics/request.md
```diff
@@ -1,316 +1,116 @@
-[//]: # (title: Request)
+[//]: # (title: Send a request)
```
hhariri, line 1: Title should be Making Requests

topics/request.md
```diff
@@ -1,316 +1,116 @@
...
 
-<include src="lib.md" include-id="outdated_warning"/>
+After adding the necessary [dependencies](client.md#add-dependencies) and [creating the client](client.md#create-client), you can send HTTP requests. You can configure the following request parameters:
```
hhariri, line 3: It's not entirely clear the phrasing here about the necessary dependencies, because it feels like "request" has its own dependency. Does it make sense for every topic to mention this "After adding...."?

Also, and use of "send" should be "make"

topics/request.md
```diff
@@ -1,316 +1,116 @@
...
+* Specify an HTTP method, such as `GET`, `POST`, `PUT`, and so on.
```
hhariri, line 4: Let's enumerate al of them. Also please let's stop using the constructs of "and so on.". It doesn't give users enough info.

topics/request.md
```diff
@@ -1,316 +1,116 @@
...
+* Add headers and cookies.
+* Set the body of a request, for example, a plain text, a data object, form parameters, and so on.
```
hhariri, line 6: ...and so on.

topics/request.md
```diff
@@ -1,316 +1,116 @@
...
-```
-
-And in the case you are interested in the raw bits, you can read a `ByteArray`:
-
-```kotlin
-val channel: ByteArray = client.get("https://en.wikipedia.org/wiki/Main_Page")
-```
-
-Or get full [HttpResponse](https://api.ktor.io/%ktor_version%/io.ktor.client.statement/-http-response/index.html):
+The main way for making HTTP requests is the [request](https://api.ktor.io/%ktor_version%/io.ktor.client.request/request.html) function that takes a URL as a parameter. Inside this function, you can configure various request parameters: specify an HTTP method, add headers, specify the request body and content type, and so on. These parameters are exposed by the [HttpRequestBuilder](https://api.ktor.io/%ktor_version%/io.ktor.client.request/-http-request-builder/index.html) class.
```
hhariri, line 8: The first sentence feels like a repeat of the above, which leads me to think that the above section can be removed entirely.

topics/request.md
```diff
@@ -1,316 +1,116 @@
...
 
 ```kotlin
-val response: HttpResponse = client.get("https://en.wikipedia.org/wiki/Main_Page")
 ```
+{src="snippets/_misc_client/RequestMethodWithoutParams.kt" interpolate-variables="true" disable-links="false"}
 
-The [HttpResponse](https://api.ktor.io/%ktor_version%/io.ktor.client.statement/-http-response/index.html) is downloaded in memory by default. To learn how to download response partially or work with a stream data consult with the [Streaming](streaming.md) section.
+Note that this function allows you to [receive a response](response.md) in various ways. In this example, we receive a response as an `HttpResponse` object.
```
hhariri, line 14: various  => mutiple

topics/request.md
```diff
@@ -1,316 +1,116 @@
...
-```kotlin
-@Serializable
-data class User(val id: Int)
 
-val response: User = client.get("https://myapi.com/user?id=1")
-```
-
-Please note that some of response types are `Closeable` and can hold resources.
+## Set request parameters {id="parameters"}
+In this section, we'll see on how to specify various request parameters, including an HTTP method, headers, and cookies. If you need to configure some default parameters for all requests of a specific client, use the [](default-request.md) feature.
```
hhariri, line 20: If you need to configure some default params should probably be it's own heading.

topics/request.md
```diff
@@ -1,316 +1,116 @@
...
-    // Get the content of an URL.
-    val firstBytes = client.get<ByteArray>("https://127.0.0.1:8080/a")
-
-    // Once the previous request is done, get the content of an URL.
-    val secondBytes = client.get<ByteArray>("https://127.0.0.1:8080/b")
-
-    client.close()
-}
 ```
+{src="snippets/_misc_client/RequestMethodWithParams.kt"}
```
hhariri, line 30: and so on...

topics/request.md
```diff
@@ -1,316 +1,116 @@
...
 
-```kotlin
-val text = client.post<String>("http://127.0.0.1:8080/") {
-    header("Hello", "World")
-}
-```
 
-The [HttpRequestBuilder](https://api.ktor.io/%ktor_version%/io.ktor.client.request/-http-request-builder/) looks like this:
+### Cookies {id="cookies"}
+To send cookies, use the [cookie](https://api.ktor.io/%ktor_version%/io.ktor.client.request/cookie.html) function:
```
hhariri, line 45: We use a lot "as follows". Not sure it's necessary. Here you don't use it, but in above examples you do. Maybe remove?

topics/request.md
```diff
@@ -1,316 +1,116 @@
...
-### Posting forms
 
-{id="submit-form"}
 
-There are a couple of convenience extension methods for submitting form information.
-The detailed reference is listed [here](https://api.ktor.io/%ktor_version%/io.ktor.client.request.forms/).
 
-The `submitForm` method:
+## Set request body {id="body"}
+To set the body of a request, you need to specify the [body](https://api.ktor.io/%ktor_version%/io.ktor.client.request/-http-request-builder/body.html) property exposed by `HttpRequestBuilder`. This property accepts different types of payload, including plain text, arbitrary class instances, form data, byte arrays, and so on. Below we'll take a look at several examples.
```
hhariri, line 57: I'd not use verb exposed here. I'd just say property via ..

payload > payloads

topics/request.md
```diff
@@ -1,316 +1,116 @@
...
-* `body = ByteArrayContent("HELLO WORLD!".toByteArray(Charsets.UTF_8))`
-* `body = LocalFileContent(File("build.gradle"))`
-* `body = JarFileContent(File("myjar.jar"), "test.txt", ContentType.fromFileExtension("txt").first())`
-* `body = URIFileContent("https://en.wikipedia.org/wiki/Main_Page")`
-
-If you install the [JsonFeature](json-feature.md), and set the content type to `application/json`
-you can use arbitrary instances as the `body`, and they will be serialized as JSON:
+## Parallel requests {id="parallel_requests"}
 
+When sending two requests at once, the client suspends the second request execution until the first is finished. If you need to perform several requests at once, you can use [launch](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/launch.html) or [async](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/async.html) functions. The code snippet below shows how to perform two requests asynchronously:
```
hhariri, line 101: first -> first one

topics/request.md
```diff
@@ -1,316 +1,116 @@
...
-client.post<Unit>() {
-    url("http://127.0.0.1:8080/")
-    body = json.write(HelloWorld(hello = "world")) // Generates an OutgoingContent
-}
-```
 
-Or using Jackson (JVM only):
+## Cancel a request {id="cancel-request"}
 
+If you need to cancel a request, you can cancel a coroutine that runs this request. For instance, the [launch](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/launch.html) function returns a `Job` that can be used to cancel the running coroutine:
```
hhariri, line 111: To cancel a request, cancel the coroutine that makes the request.

Not sure if "For instance" is needed here in this context.

topics/request.md
```diff
@@ -1,316 +1,116 @@
...
-
-```kotlin
-val request = client.request {
-    method = HttpMethod.Post
-    body = MultiPartFormDataContent(formData {
-        append("key", "value")
-    })
-}
-```
+Learn more from the [Cancellation and timeouts](https://kotlinlang.org/docs/cancellation-and-timeouts.html) help topic.
```
hhariri, line 116: Let's try and avoid using the construct of "help topic". We can just say "Learn more about Cancelations and timeouts"

hhariri (changes requested):

topics/request.md
andreyaksenov: I just want to emphasize that a user needs to set up the client before using it. Maybe, it's  better to rephrase to something like: 'After you've set up the client' (with a link to 'Setting up a client' topic).

topics/request.md
andreyaksenov: ok

topics/request.md
andreyaksenov: agree, rephrased the intro

topics/request.md
andreyaksenov: ok, removed and rephased

topics/request.md
andreyaksenov: ok, removed

topics/request.md
andreyaksenov: ok

PR #113, opened 2021-04-29
Google OAuth example

andreyaksenov:
https://github.com/ktorio/ktor-documentation/tree/auth-examples/codeSnippets/snippets/auth-oauth-google


codeSnippets/snippets/auth-oauth-google/src/main/kotlin/com/example/Application.kt
```diff
@@ -0,0 +1,63 @@
...
+    install(Authentication) {
+        oauth("auth-oauth-google") {
+            urlProvider = { "http://localhost:8080/callback" }
+            providerLookup = {
+                OAuthServerSettings.OAuth2ServerSettings(
+                    name = "google",
+                    authorizeUrl = "https://accounts.google.com/o/oauth2/auth",
+                    accessTokenUrl = "https://accounts.google.com/o/oauth2/token",
+                    requestMethod = HttpMethod.Post,
+                    clientId = System.getenv("GOOGLE_CLIENT_ID"),
```
Stexxe, line 26: We should err If environment variable isn't set.

codeSnippets/snippets/auth-oauth-google/src/main/kotlin/com/example/Application.kt
```diff
@@ -0,0 +1,63 @@
...
+                    clientSecret = System.getenv("GOOGLE_CLIENT_SECRET"),
+                    defaultScopes = listOf("https://www.googleapis.com/auth/userinfo.profile")
+                )
+            }
+            client = httpClient
+        }
+    }
+    routing {
+        var principal: OAuthAccessTokenResponse.OAuth2? = null
```
Stexxe, line 35: That's not a right way to do it. You should save an access token in a user's session.

codeSnippets/snippets/auth-oauth-google/src/main/kotlin/com/example/Application.kt
```diff
@@ -0,0 +1,63 @@
...
+        authenticate("auth-oauth-google") {
+            get("/callback") {
+                principal = call.authentication.principal()
+                call.respondRedirect("/hello")
+            }
+        }
+        get("/") {
+            call.respondRedirect("/callback")
```
Stexxe, line 43: Why do you need this redirect?

codeSnippets/snippets/auth-oauth-google/src/main/kotlin/com/example/Application.kt
```diff
@@ -0,0 +1,63 @@
...
+            }
+            call.respondText("Hello, ${userInfo.name}!")
+        }
+    }
+}
+
+data class UserInfo(
+    val id: String,
+    val name: String,
+    val given_name: String,
```
Stexxe, line 59: Code style

Stexxe (changes requested):

codeSnippets/snippets/auth-oauth-google/src/main/kotlin/com/example/Application.kt
andreyaksenov: to quickly open the Google authorization page when a user runs an application and clicks `http://0.0.0.0:8080`

codeSnippets/snippets/auth-oauth-google/src/main/kotlin/com/example/Application.kt
andreyaksenov: ok, saved a token to cookies

codeSnippets/snippets/auth-oauth-google/src/main/kotlin/com/example/Application.kt
andreyaksenov: I agree that such situations should be handled but I'd not add additional checks here to simplify a code snippet in docs. In addition, an application raises a more or less obvious exception in this case (`System.getenv("GOOGLE_CLIENT_ID") must not be null`).

codeSnippets/snippets/auth-oauth-google/src/main/kotlin/com/example/Application.kt
```diff
@@ -0,0 +1,62 @@
...
+                    append(HttpHeaders.Authorization, "Bearer ${userSession?.token}")
+                }
+            }
+            call.respondText("Hello, ${userInfo.name}!")
+        }
+    }
+}
+
+data class UserSession(val token: String)
+data class UserInfo(val id: String, val name: String, val given_name: String, val family_name: String, val picture: String, val locale: String)
```
Stexxe, line 62: The indentation was good, I meant `given_name` and `family_name` identifiers :)

codeSnippets/snippets/auth-oauth-google/src/main/kotlin/com/example/Application.kt
Stexxe: It's kinda confusing for a reader that you use `/callback` route to authenticate and as a callback URL. I'd recommend having a separate `/login` route in the `authenticate` block and redirect a user there.

codeSnippets/snippets/auth-oauth-google/src/main/kotlin/com/example/Application.kt
```diff
@@ -0,0 +1,62 @@
...
+                val principal: OAuthAccessTokenResponse.OAuth2? = call.authentication.principal()
+                call.sessions.set(UserSession(principal?.accessToken.toString()))
+                call.respondRedirect("/hello")
+            }
+        }
+        get("/") {
+            call.respondRedirect("/callback")
+        }
+        get("/hello") {
+            val userSession: UserSession? = call.sessions.get<UserSession>()
```
Stexxe, line 50: What will happen if a user doesn't have a valid session with a token?

Stexxe (changes requested):

Stexxe (approved):

PR #123, opened 2021-05-20
2.0.0 dynamic features

andreyaksenov:


andreyaksenov:
EAP docs: https://ktor.io/docs/eap/plugins.html#install-route

PR #136, opened 2021-06-29
Client response validation

andreyaksenov:
Updated the 'Response validation' topic and added two runnable examples:
- https://github.com/ktorio/ktor-documentation/tree/client-response-validation/codeSnippets/snippets/client-validate-response
- https://github.com/ktorio/ktor-documentation/tree/client-response-validation/codeSnippets/snippets/client-customize-default-validation


codeSnippets/snippets/client-customize-default-validation/src/main/kotlin/com/example/Application.kt
```diff
@@ -0,0 +1,42 @@
...
+            "Status: ${response.status}."
+}
+
+fun main() {
+    val client = HttpClient(CIO) {
+        HttpResponseValidator {
+            handleResponseException { exception ->
+                if (exception !is ClientRequestException) {
+                    return@handleResponseException
+                } else {
```
rsinukov, line 23: no need for `else` branch, since `if` has `return`

rsinukov (approved):

PR #149, opened 2021-07-14
Document that Ktor clients and servers support Android 5+

gnarea:
@relaycorp has been running ktor on Android 5+ devices for a while now. Both [clients](https://github.com/relaycorp/awala-poweb-jvm) and servers, and both backed by Netty.

Except for [this minor regression recently introduced in ktor 1.6.1](https://youtrack.jetbrains.com/issue/KTOR-2924), ktor has been running fine on Android 5+ for us.

### DIFF
--- a/topics/FAQ.md
+++ b/topics/FAQ.md
@@ -146,23 +146,9 @@ to have a working project as base.
 ## Can I use ktor on Android?
 {id="android-support"}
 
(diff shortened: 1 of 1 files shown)


andreyaksenov:
@gnarea thanks for the contribution! Information added to the upcoming 2.0.0 docs: https://ktor.io/docs/eap/faq.html

PR #152, opened 2021-07-22
Adding documentation for "wildcard" static file routing

EyeOfMidas:
This is mentioned offhandedly in https://github.com/ktorio/ktor/issues/514 as a workaround solution.
This behavior is not called out in the docs anywhere and is very useful to know.

### DIFF
--- a/topics/Serving_Static_Content.md
+++ b/topics/Serving_Static_Content.md
@@ -68,6 +68,17 @@ routing {
 }
 ```
 
+### Serving all files including in subfolders
+The `file` function can also take the string `"."` which will serve up any file as long as the request path and physical filename match. Ktor will use the filename extension to determine the `Content-Type` of the file requested.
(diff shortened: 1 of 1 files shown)


andreyaksenov:
@EyeOfMidas thanks! Merging

Virtlink:
The documentation now mentions `file(".")`, which does not work. It should be `files(".")`, note the plural, as mentioned in [the workaround](https://github.com/ktorio/ktor/issues/514#issuecomment-586699284).

andreyaksenov:
> The documentation now mentions `file(".")`, which does not work. It should be `files(".")`, note the plural, as mentioned in [the workaround](https://github.com/ktorio/ktor/issues/514#issuecomment-586699284).

@Virtlink thanks for noticing, fixed in 1.6.7 docs.

PR #167, opened 2021-08-24
staticBasePackage takes a string instead of File

EyeOfMidas:
Adjusting the documentation to show how to use this properly

https://api.ktor.io/ktor-server/ktor-server-core/ktor-server-core/io.ktor.http.content/static-base-package.html


andreyaksenov:
@EyeOfMidas thanks for the contribution! This issue is fixed in the updated topic: https://ktor.io/docs/eap/serving-static-content.html#default-resource-package

PR #174, opened 2021-09-05
Implementation of sumOf instead of map

yuribreion1:
With the implementation of sumOf, the code seems to have more semantics.


andreyaksenov:
@yuribreion1 thanks for the contribution! In the EAP 2.0.0 docs, `map` is replaced with `sumOf `: https://ktor.io/docs/eap/creating-http-apis.html#totalize_order

PR #177, opened 2021-09-08
Using more incluside words for allowed methods

aleixmorgadas:
I renamed the `white list` to `allow list` to refer to the allowed HTTP CORS Methods.


andreyaksenov:
@aleixmorgadas Thanks! Merging

PR #186, opened 2021-11-23
change "artifact" to "setArtifact"

programmerkgit:
`artifact = "..."` was not supported. 

setArtifact("...") works.

### DIFF
--- a/codeSnippets/snippets/google-appengine-standard/build.gradle
+++ b/codeSnippets/snippets/google-appengine-standard/build.gradle
@@ -15,7 +15,7 @@ repositories {
 
 appengine {
     stage {
-        artifact = "build/libs/${project.name}-${project.version}-all.jar"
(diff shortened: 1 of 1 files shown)


andreyaksenov:
@programmerkgit thank you for the contribution. In the upcoming 2.0.0 docs, we've updated all Gradle scripts from Groovy to Kotlin, so the Google App Engine sample already uses `setArtifact`: https://ktor.io/docs/eap/google-app-engine.html#configure-app-engine-plugin (step 3)

programmerkgit:
> @programmerkgit thank you for the contribution. In the upcoming 2.0.0 docs, we've updated all Gradle scripts from Groovy to Kotlin, so the Google App Engine sample already uses `setArtifact`: https://ktor.io/docs/eap/google-app-engine.html#configure-app-engine-plugin (step 3)

Thank you for taking time to reply.  
I love ktor. 応援しています！

PR #199, opened 2021-12-25
Update headlines for Ktor Client / Json

SebastianAigner:
This PR updates the the Ktor Clients "Json" page:

- `kotlinx.serialization` is referred to in full everywhere (`kotlinx` is not an abbreviation for `kotlinx.serialization` – we have about 50 other libraries starting with `kotlinx` :) )
- Headlines are restructured to first show the name of the library, and then the platforms for which they are available.


andreyaksenov:
Thanks! Merging

PR #203, opened 2021-12-30
Fix import location

kurly09:


andreyaksenov:
@kurly09 thanks for the feedback. Documentation for the 2.0 version lives in a separate `2.0.0` branch and is published on https://ktor.io/docs/eap/. In [this branch](https://github.com/ktorio/ktor-documentation/tree/2.0.0), all code snippets are migrated to a new version and use new imports.

PR #208, opened 2022-01-15
Update ConversionService and DelegatingConversionService examples with their new signature

ddadon10:
Update the examples of [ConversionService](https://github.com/ktorio/ktor/blob/0de7948fbe3f78673f4f90de9c5ea5986691819a/ktor-utils/common/src/io/ktor/util/converters/ConversionService.kt#L13) and [DelegatingConversionService](https://github.com/ktorio/ktor/blob/99d172d9f8df1e95c58d4335d599ff20ac0d29e6/ktor-utils/common/src/io/ktor/util/converters/DataConversion.kt#L73) so they match their signatures.

### DIFF
--- a/topics/data-conversion.md
+++ b/topics/data-conversion.md
@@ -102,16 +102,18 @@ val dataConversion = call.conversionService
 
 ```kotlin
 interface ConversionService {
-    fun fromValues(values: List<String>, type: Type): Any?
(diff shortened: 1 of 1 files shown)


andreyaksenov:
@ddadon10 thanks for the contribution! Merging

PR #219, opened 2022-01-30
Update kotlin_serialization page to contain 2.0.0 package

Neitex:
`ktor-serialization` package does not contain `json()` function, but `ktor-serialization-kotlinx-json` does (as far as I understand, that changed with Ktor 2.0.0)


Neitex:
Just noticed that there are version selector at top of page :)

PR #224, opened 2022-02-16
Make native server limitations a bit more clear

jakobkmar:
- make sure to mention the reverse proxy option for HTTPS
- mention that only the windows target is not supported, you can still develop on windows


andreyaksenov:
@jakobkmar thanks! Merging

PR #229, opened 2022-03-19
optimize: add jwt challenge handle case

chanfan1999:
optimize: add jwt challenge handle case


andreyaksenov:
@chanfan1999 thanks for your contribution! I've added info about `challenge` to the [JSON Web Tokens](https://ktor.io/docs/eap/jwt.html#validate-payload) topic and updated corresponding samples:
- [auth-jwt-hs256](https://github.com/ktorio/ktor-documentation/blob/2.0.0/codeSnippets/snippets/auth-jwt-hs256/src/main/kotlin/com/example/Application.kt)
- [auth-jwt-rs256](https://github.com/ktorio/ktor-documentation/blob/2.0.0/codeSnippets/snippets/auth-jwt-rs256/src/main/kotlin/com/example/Application.kt)

PR #233, opened 2022-04-08
Fix plugin names

turansky:


andreyaksenov:
@turansky thanks for the fix! Merging

PR #234, opened 2022-04-12
Update ApplicationTest.kt

victororlyk:
add application config, so the test does not fail. 
Without it, for me test fails and throws errors, if creating a project from scratch following this tutorial 
https://ktor.io/docs/creating-http-apis.html

### DIFF
--- a/codeSnippets/snippets/tutorial-http-api/src/test/kotlin/com/example/ApplicationTest.kt
+++ b/codeSnippets/snippets/tutorial-http-api/src/test/kotlin/com/example/ApplicationTest.kt
@@ -5,8 +5,14 @@ import io.ktor.client.statement.*
 import io.ktor.http.*
 import io.ktor.server.testing.*
 import kotlin.test.*
+import com.example.plugins.*
 
 class OrderRouteTests {
+    application{
+        configureRouting()
+        configureSerialization()
+        configureTemplating()
+    }
     @Test
     fun testGetOrder() = testApplication {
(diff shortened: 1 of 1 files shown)


andreyaksenov:
@victororlyk did you changed the `Configuration in` option to `HOCON file` when creating a project as described here https://ktor.io/docs/creating-http-apis.html#create_ktor_project? In this case, `testApplication` should load modules specified in the `application.conf` file automatically (https://ktor.io/docs/testing.html#add-modules).

andreyaksenov:
@victororlyk yep, the `application { configureRouting() }` code is added regardless of the selected configuration option but it's not required when HOCON is used because modules are added automatically. I'll add a corresponding note in the [Automated testing﻿](https://ktor.io/docs/creating-http-apis.html#automated_testing) chapter.

victororlyk:
yeah, for me as a newbie in Kotlin, it was not clear that  is why wanted to help with  documentation. 

When  I see what plugin generates against what is writen in learning notes ( the difference in that application... thing) it was making me think something is wrong. 
And if not choosing hocon option it fails. So thought it will not harm to add that part)

andreyaksenov:
@victororlyk ok, I agree that it's better to have `application { ... }` and show how to load modules explicitly to avoid such issues when going through a tutorial. I'll merge you PR but will remove the `configureTemplating()` module because it's missing in a tutorial. Thanks for your contribution!

PR #240, opened 2022-04-30
Inline explanations for request, connect, and socket timeouts

SebastianAigner:
https://youtrack.jetbrains.com/issue/KTOR-4261/Explanation-on-request,-connection,-and-socket-missing-in-KDoc

"`requestTimeoutMillis` for a request timeout." does not add the valuable information of what a _request timeout_ even means – it requires readers to manually cross-reference to the listing at the top of the page, which seems like something that could easily be avoided by including the information right at this place. This PR addresses this.


andreyaksenov:
@SebastianAigner thanks! Merging

PR #249, opened 2022-06-08
Make two small corrections

MasoodFallahpoor:


andreyaksenov:
@MasoodFallahpoor thanks!

PR #250, opened 2022-06-24
Update double-receive.md regarding cacheRawRequest

pro-ship-it:
The explanation of `cacheRawRequest` seems misleading. Its default value is `true`, which already allows to receive different types from the same request.


andreyaksenov:
@pro-ship-it thanks for the contribution! Merging

PR #253, opened 2022-06-30
change `AtomicInteger` property from `var` to `val`

aSemy:
`lastId` itself isn't reassigned - the integer inside `AtomicInteger` is mutated. Therefore `lastId` can be a `val`. This makes it clearer that `lastId` shouldn't be reassigned.

The Coroutines docs have an example where an `AtomicInteger` is defined using `val`

https://kotlinlang.org/docs/shared-mutable-state-and-concurrency.html#thread-safe-data-structures

```kotlin
val counter = AtomicInteger()

fun main() = runBlocking {
    withContext(Dispatchers.Default) {
        massiveRun {
            counter.incrementAndGet()
        }
    }
    println("Counter = $counter")
}
```


andreyaksenov:
@aSemy thanks for the contribution! Merging

PR #254, opened 2022-07-06
KTOR-3179 request validation

andreyaksenov:


codeSnippets/snippets/request-validation/src/main/kotlin/com/example/Application.kt
```diff
@@ -0,0 +1,66 @@
...
+import kotlinx.serialization.*
+
+@Serializable
+data class Customer(val id: Int, val firstName: String, val lastName: String)
+
+fun main(args: Array<String>): Unit = io.ktor.server.netty.EngineMain.main(args)
+
+fun Application.module() {
+    install(RequestValidation) {
+        validate<String> { bodyText ->
```
Stexxe, line 21: These validations are artificial and in most cases won't be encountered in real-world applications.

PR #260, opened 2022-09-02
Fix Netty HTTP/2 samples

rsinukov:
3 things were broken:
* `ktor-netty` itself. I will open PR in the main repo shortly
* having both `io.netty:netty-tcnative` and `io.netty:netty-tcnative-boringssl-static`. You need to add only one of them (preferably BoringSSL), otherwise Netty fails with duplicate symbol exception
* version of Netty and TCNative should be in sync, since Netty treats TCNative as the internal dependency and relies on the behaviour. 

@andreyaksenov, Do you mind updating tutorials considering points2 and 3?


andreyaksenov:
@rsinukov thanks, I've applied your changes and updated the topic: https://ktor.io/docs/advanced-http2.html#netty

PR #262, opened 2022-09-12
Supply typical development mode configuration for Thymeleaf

delitescere:
This allows for "hot reloading" of uncached template files from the project source directory while in development mode.


andreyaksenov:
@delitescere thanks for your contribution! I've extracted your sample into a separate project and referenced it in docs: https://ktor.io/docs/thymeleaf.html#auto-reload.

PR #266, opened 2022-10-05
Update getting_started_ktor_client.topic

inez:
`main` must be a suspend function if it calls another (`get`) suspend function


andreyaksenov:
@inez thanks for your contribution. Step 4 in the [Create a client﻿](https://ktor.io/docs/getting-started-ktor-client.html#create-client) already shows how to make `main` suspend.

PR #276, opened 2022-11-17
KTOR-5173 new custom plugins API

andreyaksenov:


topics/client-custom-plugins.md
```diff
@@ -0,0 +1,269 @@
...
+<a href="https://api.ktor.io/ktor-http/io.ktor.http.content/-outgoing-content/index.html">OutgoingContent</a> 
+(for example, <code>TextContent</code>, <code>ByteArrayContent</code>, or <code>FormDataContent</code>)
+or return <code>null</code> if your transformation is not applicable.
+</td>
+</tr>
+
+
+<tr>
+<td>
+<code>on(Send)</code>
```
rsinukov, line 136: What do you think about separating basic callbacks from hooks? Up to you, but I think it may be easier for people to understand first basic API and then learn about hooks.

rsinukov (approved):

topics/client-custom-plugins.md
andreyaksenov: Good point, thanks. I thought about it but still didn't find a good solution to structure the topic in this case. I'll file a corresponding issue and will make another try a bit later.

PR #278, opened 2022-11-25
Change wrong parameter in extensions example

nartsisss:
`MyWebSocketException` should be named `MyWebSocketExtension` as in the previous example.


andreyaksenov:
Thanks for the contribution! Merging.

PR #284, opened 2022-12-10
Update google oauth guide with a solution for the `redirect_uri_mismatch` issue.

Kietyo:
With some nice screenshots as well.


andreyaksenov:
Thanks for your contribution! We've added the dedicated section about creating and configuring authorization credentials:
https://ktor.io/docs/oauth.html#authorization-credentials

This section also mentions that the route specified in `urlProvider ` should be added to a list of Authorised redirect URIs.
https://ktor.io/docs/oauth.html#configure-oauth-provider

PR #287, opened 2022-12-27
Explain DSL and their limitations

cies:
We've been bitten by the VERY slow compilation speed of these DSLs. It really messes with our work-flow and hence I think it is nice to prevent others from making the same mistake.


vnikolova:
The compiling speed issues have been fixed in kotlinx.html  (see [Release 0.9.0](https://github.com/Kotlin/kotlinx.html/releases/tag/0.9.0) and [0.9.1](https://github.com/Kotlin/kotlinx.html/releases/tag/0.9.1))

PR #292, opened 2023-02-14
KTOR-5567 Documentation for regex patterns in routing

marychatte:
[Design](https://youtrack.jetbrains.com/issue/KTOR-5279/Design-for-regex-patterns-support-in-routing)

### DIFF
--- a/topics/Routing_in_Ktor.md
+++ b/topics/Routing_in_Ktor.md
@@ -126,7 +126,11 @@ To access path segments' values inside the route handler, use `call.parameters.g
 
 ### Regular expression {id="regular_expression"}
 
-Regular expressions can be used with all defining route handlers functions: `route`, `get`, `post`, and so on. For example:
+Regular expressions can be used with all defining route handlers functions: `route`, `get`, `post`, and so on. 
+
+> To know more about regular expressions, see [Kotlin documentation](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/-regex/).
+
+Let's write a route matches any path that ends with `/hello`.
 
 ```kotlin
 import io.ktor.server.routing.*
@@ -138,11 +142,19 @@ routing {
     }
 }
 ```
+With this route definition, any incoming request to a path ending with `/hello`, such as `/foo/hello`, `/bar/baz/hello`, 
+and so on, will be matched.
 
-#### Named groups
+#### Accessing path parts in handler
 
-To declare a named group in the regular expression, use `(?<name>...)` syntax.
-You can use the `call.parameters` property to access a named group inside the route handler. For example, `call.parameters["id"]` in the code snippet below will return `123` for the `123/hello` path:
+In regular expressions, named groups are a way to capture a specific part of a string that matches a pattern and assign it a name.
+Named groups are defined using the following syntax `(?<name>pattern)`, where `name` is the name of the group and
+`pattern` is the regular expression pattern that matches the group.
+
+By defining a named group in the route function, you can capture a part of the path, and then in the handler function,
+you can access the captured parameter using the `call.parameters` property.
+
+For example, you can define a route that matches requests to a path that includes an integer identifier followed by `/hello`.
 
 ```kotlin
 import io.ktor.server.routing.*
@@ -155,6 +167,9 @@ routing {
     }
 }
 ```
+In the code below, the `(?<id>\\d+)` named group is used to capture the integer identifier `id` from the path,
+and the `call.parameters` property is used to access the captured id parameter in the handler function.
+
 Unnamed groups can't be accessed inside the route handler, but you can use them to match the path. For example, path `hello/world` will be matched while `hello/World` not:
 
 ```kotlin
@@ -167,10 +182,8 @@ routing {
     }
 }
 ```
-Also, if the path matches by regex, it should start with a forward slash or nothing after the part consumed by regex. For example, path pattern `get(Regex("[a-z]+"))` will not match the path "hello1" 
-but will match the part `hello` of the path `hello/1` and leave `/1` for the next route. 
-
-> To know more about regular expressions, see [Kotlin documentation](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/-regex/).
+Also, the whole path segment should be consumed by regex. For example, path pattern `get(Regex("[a-z]+"))` will not match the path `"hello1"` 
+but will match the part `hello` of the path `hello/1` and leave `/1` for the next route.
 
 ## Define multiple route handlers {id="multiple_routes"}
 
--- a/topics/Routing_in_Ktor.md
+++ b/topics/Routing_in_Ktor.md
@@ -130,7 +130,7 @@ Regular expressions can be used with all defining route handlers functions: `rou
 
 > To know more about regular expressions, see [Kotlin documentation](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/-regex/).
 
-Let's write a route matches any path that ends with `/hello`.
+Let's write a route that matches any path that ends with `/hello`.
 
 ```kotlin
 import io.ktor.server.routing.*
@@ -148,11 +148,11 @@ and so on, will be matched.
 #### Accessing path parts in handler
 
 In regular expressions, named groups are a way to capture a specific part of a string that matches a pattern and assign it a name.
-Named groups are defined using the following syntax `(?<name>pattern)`, where `name` is the name of the group and
-`pattern` is the regular expression pattern that matches the group.
+The syntax `(?<name>pattern)` is used to define named groups, where `name` is the name of a group and 
+`pattern` is a regular expression pattern that matches the group.
 
-By defining a named group in the route function, you can capture a part of the path, and then in the handler function,
-you can access the captured parameter using the `call.parameters` property.
+By defining a named group in a route function, you can capture a part of the path, and then in the handler function,
+you can access the captured parameter using the `call.parameters` object.
 
 For example, you can define a route that matches requests to a path that includes an integer identifier followed by `/hello`.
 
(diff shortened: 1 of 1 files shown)


topics/Routing_in_Ktor.md
```diff
@@ -122,6 +124,54 @@ To access a parameter value inside the route handler, use the `call.parameters`
...
+import io.ktor.server.routing.*
+import io.ktor.server.response.*
+
+routing {
+    get(Regex("hello/[a-z]+")) {
+        call.respondText("Hello")
+    }
+}
+```
+Also, if the path matches by regex, it should start with a forward slash or nothing after the part consumed by regex. For example, path pattern `get(Regex("[a-z]+"))` will not match the path "hello1"
```
rsinukov, line 170: It was difficult for me to understand the meaning here. Maybe rephrase it, saying that the whole path segment should be consumed?

rsinukov (changes requested):

topics/Routing_in_Ktor.md
marychatte: Thanks, so I will change it for the next version:
```
Also, the whole path segment should be consumed by regex. For example, path pattern `get(Regex("[a-z]+"))` will not match the path `"hello1"` 
but will match the part `hello` of the path `hello/1` and leave `/1` for the next route.
```
Is it okay?

topics/Routing_in_Ktor.md
```diff
@@ -122,6 +124,54 @@ To access a parameter value inside the route handler, use the `call.parameters`
...
+routing {
+    get(Regex(".+/hello")) {
+        call.respondText("Hello")
+    }
+}
+```
+
+#### Named groups
+
+To declare a named group in the regular expression, use `(?<name>...)` syntax.
```
e5l, line 144: please define what the named group is and why it is needed

topics/Routing_in_Ktor.md
```diff
@@ -122,6 +124,54 @@ To access a parameter value inside the route handler, use the `call.parameters`
 A path parameter with a tailcard (`{param...}`) matches all the rest of the URL path and puts multiple values for each path segment into parameters using `param` as a key. For example, `/user/{param...}` matches `/user/john/settings`.
 To access path segments' values inside the route handler, use `call.parameters.getAll("param")`. For the example above, the `getAll` function will return an array containing the _john_ and _settings_ values.
 
+### Regular expression {id="regular_expression"}
+
+Regular expressions can be used with all defining route handlers functions: `route`, `get`, `post`, and so on. For example:
```
e5l, line 129: It would be great to add a link to the syntax of used regular expressions directly here.

topics/Routing_in_Ktor.md
```diff
@@ -122,6 +124,54 @@ To access a parameter value inside the route handler, use the `call.parameters`
...
+
```
e5l, line 130: Please define a problem which you're going to solve with a sample: `Let's write a route matches blahblah`

topics/Routing_in_Ktor.md
```diff
@@ -122,6 +124,54 @@ To access a parameter value inside the route handler, use the `call.parameters`
...
+import io.ktor.server.response.*
+
+routing {
+    get(Regex(".+/hello")) {
+        call.respondText("Hello")
+    }
+}
+```
+
+#### Named groups
```
e5l, line 142: I would name it using the purpose: `Accessing path parts in handler` and next explain how to use groups for that

e5l (changes requested):

marychatte:
@rsinukov and @e5l, thanks for your comments. I tried to fix all the problems. Could you please recheck it?

topics/Routing_in_Ktor.md
```diff
@@ -122,6 +124,67 @@ To access a parameter value inside the route handler, use the `call.parameters`
...
+        call.respondText("Hello")
+    }
+}
+```
+With this route definition, any incoming request to a path ending with `/hello`, such as `/foo/hello`, `/bar/baz/hello`, 
+and so on, will be matched.
+
+#### Accessing path parts in handler
+
+In regular expressions, named groups are a way to capture a specific part of a string that matches a pattern and assign it a name.
```
e5l, line 150: is `,` needed?

e5l (approved):
Looks better, lgtm

topics/Routing_in_Ktor.md
marychatte: It is what `Grammarly` recommends to do :)
As I understand, you can add a comma or not for short introductory phrases

e5l: got it, thanks

topics/Routing_in_Ktor.md
```diff
@@ -122,6 +124,67 @@ To access a parameter value inside the route handler, use the `call.parameters`
 A path parameter with a tailcard (`{param...}`) matches all the rest of the URL path and puts multiple values for each path segment into parameters using `param` as a key. For example, `/user/{param...}` matches `/user/john/settings`.
 To access path segments' values inside the route handler, use `call.parameters.getAll("param")`. For the example above, the `getAll` function will return an array containing the _john_ and _settings_ values.
 
+### Regular expression {id="regular_expression"}
+
+Regular expressions can be used with all defining route handlers functions: `route`, `get`, `post`, and so on. 
+
+> To know more about regular expressions, see [Kotlin documentation](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/-regex/).
+
+Let's write a route matches any path that ends with `/hello`.
```
Stexxe, line 133: **that** matches

topics/Routing_in_Ktor.md
```diff
@@ -122,6 +124,67 @@ To access a parameter value inside the route handler, use the `call.parameters`
...
+
+```kotlin
+import io.ktor.server.routing.*
+import io.ktor.server.response.*
+
+routing {
+    get(Regex(".+/hello")) {
```
Stexxe, line 140: It would be helpful to mention that a regex of a route is implicitly surrounded by `^` and `$`.

topics/Routing_in_Ktor.md
```diff
@@ -122,6 +124,67 @@ To access a parameter value inside the route handler, use the `call.parameters`
...
+}
+```
+With this route definition, any incoming request to a path ending with `/hello`, such as `/foo/hello`, `/bar/baz/hello`, 
+and so on, will be matched.
+
+#### Accessing path parts in handler
+
+In regular expressions, named groups are a way to capture a specific part of a string that matches a pattern and assign it a name.
+Named groups are defined using the following syntax `(?<name>pattern)`, where `name` is the name of the group and
+`pattern` is the regular expression pattern that matches the group.
```
Stexxe, line 152: * using the `(?<name>pattern)` syntax
* is the name of **a** group
* `pattern` is **a** regular expression pattern that matches the group

topics/Routing_in_Ktor.md
```diff
@@ -122,6 +124,67 @@ To access a parameter value inside the route handler, use the `call.parameters`
...
+
+By defining a named group in the route function, you can capture a part of the path, and then in the handler function,
```
Stexxe, line 154: in *a* route function

topics/Routing_in_Ktor.md
```diff
@@ -122,6 +124,67 @@ To access a parameter value inside the route handler, use the `call.parameters`
...
+you can access the captured parameter using the `call.parameters` property.
+
+For example, you can define a route that matches requests to a path that includes an integer identifier followed by `/hello`.
+
+```kotlin
+import io.ktor.server.routing.*
+import io.ktor.server.response.*
+
+routing {
+    get(Regex("(?<id>\\d+)/hello")) {
```
Stexxe, line 164: Maybe we should use triple quotes in the examples to avoid double backslashes.

topics/Routing_in_Ktor.md
```diff
@@ -122,6 +124,67 @@ To access a parameter value inside the route handler, use the `call.parameters`
...
+        val id = call.parameters["id"]!!
+        call.respondText(id)
+    }
+}
+```
+In the code below, the `(?<id>\\d+)` named group is used to capture the integer identifier `id` from the path,
```
Stexxe, line 170: from a requested path

topics/Routing_in_Ktor.md
```diff
@@ -122,6 +124,67 @@ To access a parameter value inside the route handler, use the `call.parameters`
...
+and so on, will be matched.
+
+#### Accessing path parts in handler
+
+In regular expressions, named groups are a way to capture a specific part of a string that matches a pattern and assign it a name.
+Named groups are defined using the following syntax `(?<name>pattern)`, where `name` is the name of the group and
+`pattern` is the regular expression pattern that matches the group.
+
+By defining a named group in the route function, you can capture a part of the path, and then in the handler function,
+you can access the captured parameter using the `call.parameters` property.
```
Stexxe, line 155: It's better to use the word `object` instead of `property` because it doesn't matter what the relationship is between `ApplicationCall` and `parameters`.

topics/Routing_in_Ktor.md
```diff
@@ -122,6 +124,67 @@ To access a parameter value inside the route handler, use the `call.parameters`
...
+
+routing {
+    get(Regex("(?<id>\\d+)/hello")) {
+        val id = call.parameters["id"]!!
+        call.respondText(id)
+    }
+}
+```
+In the code below, the `(?<id>\\d+)` named group is used to capture the integer identifier `id` from the path,
+and the `call.parameters` property is used to access the captured id parameter in the handler function.
```
Stexxe, line 171: id in backticks

topics/Routing_in_Ktor.md
```diff
@@ -122,6 +124,67 @@ To access a parameter value inside the route handler, use the `call.parameters`
...
+
+Unnamed groups can't be accessed inside the route handler, but you can use them to match the path. For example, path `hello/world` will be matched while `hello/World` not:
```
Stexxe, line 173: Unnamed groups can't be accessed inside **a regex route handler**
There are no unnamed groups in the following example.

topics/Routing_in_Ktor.md
```diff
@@ -122,6 +124,67 @@ To access a parameter value inside the route handler, use the `call.parameters`
...
+import io.ktor.server.routing.*
+import io.ktor.server.response.*
+
+routing {
+    get(Regex("hello/[a-z]+")) {
+        call.respondText("Hello")
+    }
+}
+```
+Also, the whole path segment should be consumed by regex. For example, path pattern `get(Regex("[a-z]+"))` will not match the path `"hello1"`
```
Stexxe, line 185: What do you mean here by **should**? Is it consumed or not?

topics/Routing_in_Ktor.md
marychatte: It doesn't work for the route function. In next example `Regex(".+/hello")` wouldn't end with `$`:
```
route(Regex(".+/hello")) {
    get(Regex("/world") {
    ...
    }
}
```

PR #297, opened 2023-03-30
KTOR-5744 Documentation for Support loading multiple configuration files

rsinukov:


e5l (approved):
lgtm

e5l:
Maybe we also need some sample, but I'm not sure how to add it

PR #298, opened 2023-04-01
Update response.md to have more information

milosgarunovic:
Update response.md to have more information to have more information on JSON object section. I had an issue because I already had content negotiation for the server and I couldn't import it. So after a bit of time I found out I need to add io.ktor:ktor-client-content-negotiation in gradle to make it work. This is just more of a suggestion so it's more user friendly and to have everything in one place. I know there's link on 66th line about ContentNegotiation for client but even that didn't help in timely manner.


rsinukov (approved):
Thanks for the PR. LGTM.

PR #300, opened 2023-04-06
KTOR-5245 Documentation for Support Compression of Request Body with ContentEncoding Client Plugin

rsinukov:


e5l (changes requested):
Should we base it on top of 3.0?

rsinukov:
Thanks, fixed

e5l (approved):
lgtm

PR #305, opened 2023-04-20
Update ProGuard keep rules for Ktor 2.3.0

ccjernigan:
I was getting runtime crashes with after updating to Ktor 2.3.0, and it looks like the ProGuard rules needed a small update.


rsinukov (approved):
LGTM

PR #312, opened 2023-06-06
Database persistence with exposed - 'DatabaseFactory' object name change to 'DatabaseSingleton'

ktprezes:
Kotlin object is 'Singleton' by its nature, so I've found the name 'DatabaseFactory' as misleading - IMHO it should be changed to 'DatabaseSingleton' or 'DatabaseObject'


vnikolova:
Merged. Thank you for your contribution @ktprezes!

PR #320, opened 2023-08-05
Updated Code Snippet

phalcorine:
Wrap the text used in the `send()` method on line 25 with `Frame.Text()`


vnikolova:
Thank you for the suggestion @phalcorine. In this case, wrapping the text with `Frame.Text()` is not necessary as the `send()` method overload handles this for `String` content.

PR #325, opened 2023-09-10
Add libcurl-gnutls info to curl

StefMa:
I just playing around a bit with ktor, linux and the curl client.
After it failed a few times on the CI machine, I realized that I have to install the `libcurl-gnutls` version.
libcurl has different versions available. 
It also have the `openssl` and `nns` version.
While I can confirm that the `openssl` versions doesn't work, I don't know if it will work with the `nns` version.

Whatever, I think it make sense to document this 🙃


e5l (approved):
Please add `sudo apt install ....` command for ubuntu

PR #347, opened 2023-10-04
KTOR-6156 docs for CallId client plugin

vnikolova:
[KTOR-6156](https://youtrack.jetbrains.com/issue/KTOR-6156/Documentation-for-CallId-Feature-for-Client)
- added documentation for CallId client plugin
- created a new code snippet with example. The example uses a Ktor server to demonstrate how call ID is transferred between multiple calls
- added a link summary to the CallId server plugin doc

### DIFF
--- a/codeSnippets/snippets/client-call-id/README.md
+++ b/codeSnippets/snippets/client-call-id/README.md
@@ -21,7 +21,7 @@ The client makes 3 requests to the API and prints out information about the call
 
 The expected output is:
 ```
-in call: call-id-client, in header: call-id-client, in query param: call-id-client
-in call: call-id-client, in header: call-id-client, in query param: null
-in call: call-id-server, in header: null, in query param: call-id-client-2
+Handling request `/call` from client with call.callId: call-id-client and X-RequestId: call-id-client
+Handling request `/call` from client with call.callId: call-id-client and X-RequestId: call-id-client
+Handling request `/call` from client with call.callId: call-id-client-2 and X-RequestId: call-id-client-2
 ```
--- a/codeSnippets/snippets/client-call-id/README.md
+++ b/codeSnippets/snippets/client-call-id/README.md
@@ -1,0 +1,27 @@
+# Client call id
+
+A sample Ktor project showing how to use the [CallId](https://ktor.io/docs/client-call-id.html) client plugin.
+The project consists of two parts: an API service and a client.
+The API service utilizes the `callId` plugin on the server and configures it to retrieve a call ID from the request
+header or use a generated call ID.
+
+> This sample is a part of the [codeSnippets](../../README.md) Gradle project.
+
+## Running
+
+1. To start the API service, navigate to `CallIdService.kt` and click on the play button next to the `main()` function.
+
+2. Run the client sample by executing the following command:
+
+```bash
+./gradlew :client-call-id:run
+```
+
+The client makes 3 requests to the API and prints out information about the call ID.
+
+The expected output is:
+```
+in call: call-id-client, in header: call-id-client, in query param: call-id-client
+in call: call-id-client, in header: call-id-client, in query param: null
+in call: call-id-server, in header: null, in query param: call-id-client-2
+```
--- a/codeSnippets/snippets/client-call-id/src/main/kotlin/com/example/Application.kt
+++ b/codeSnippets/snippets/client-call-id/src/main/kotlin/com/example/Application.kt
@@ -8,32 +8,37 @@ import io.ktor.client.plugins.callid.*
 import io.ktor.http.*
 import kotlinx.coroutines.*
 
-fun main() {
-    runBlocking {
-        val client = HttpClient(CIO) {
-            install(CallId) {
-                generate { "call-id-client" }
-                addToHeader(HttpHeaders.XRequestId)
-                intercept { request, callId -> request.parameter("callId", callId) }
-            }
-        }
+suspend fun main() {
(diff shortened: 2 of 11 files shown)


codeSnippets/snippets/client-call-id/src/main/kotlin/com/example/Application.kt
```diff
@@ -0,0 +1,39 @@
...
+import io.ktor.client.statement.*
+import io.ktor.client.plugins.callid.*
+import io.ktor.http.*
+import kotlinx.coroutines.*
+
+fun main() {
+    runBlocking {
+        val client = HttpClient(CIO) {
+            install(CallId) {
+                useCoroutineContext = true
```
rsinukov, line 15: I think we should not explicitly set default values in config

codeSnippets/snippets/client-call-id/src/main/kotlin/com/example/CallIdService.kt
```diff
@@ -0,0 +1,47 @@
...
+import io.ktor.server.routing.*
+
+fun main(args: Array<String>) {
+    embeddedServer(Netty, port = 8090, host = "0.0.0.0", module = Application::module)
+        .start(wait = true)
+}
+
+fun Application.module() {
+    val client = HttpClient(CIO) {
+        install(ClientCallId) {
```
rsinukov, line 23: unnecessary `{}`

topics/client_call_id.md
```diff
@@ -0,0 +1,106 @@
...
+
+The %plugin_name% plugin configuration, provided by the [CallIdConfig]() class, allows you to generate a call ID and add
+it to the call context.
+
+### Generate a call ID
+
+Generate a call ID for a specific request in one of the following ways:
+
+* The `useCoroutineContext` property, enabled by default, adds a generator that uses the current `CoroutineContext` to
+  retrieve a call ID.
```
rsinukov, line 44: It worth mentioning that server plugin adds call id to the coroutine context

rsinukov (approved):

codeSnippets/snippets/client-call-id/src/main/kotlin/com/example/Application.kt
vnikolova: I set it to false in the next example and adjusted the docs.

codeSnippets/snippets/client-call-id/src/main/kotlin/com/example/Application.kt
```diff
@@ -0,0 +1,39 @@
...
+import io.ktor.client.engine.cio.*
+import io.ktor.client.request.*
+import io.ktor.client.statement.*
+import io.ktor.client.plugins.callid.*
+import io.ktor.http.*
+import kotlinx.coroutines.*
+
+fun main() {
+    runBlocking {
+        val client = HttpClient(CIO) {
```
e5l, line 13: I would extract every plugin usage to a separate function with a specific name

codeSnippets/snippets/client-call-id/src/main/kotlin/com/example/Application.kt
```diff
@@ -0,0 +1,39 @@
...
+            }
+        }
+
+        val response: HttpResponse = client.get("http://0.0.0.0:8090/call")
+        println(response.bodyAsText())
+
+        val responseFromDoubleCall: HttpResponse = client.get("http://0.0.0.0:8090/double-call")
+        println(responseFromDoubleCall.bodyAsText())
+
+        val client2 = HttpClient(CIO) {
```
e5l, line 27: client2 should also be closed.

I would write it using `use` extension: 
```
client2.use {
}
```

codeSnippets/snippets/client-call-id/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+```bash
+./gradlew :client-call-id:run
+```
+
+The client makes 3 requests to the API and prints out information about the call ID.
+
+The expected output is:
+```
+in call: call-id-client, in header: call-id-client, in query param: call-id-client
```
e5l, line 24: Could we print it in a way that indicates how the data flow goes? For instance, in a server client we can use call id from the request as a part, so we can track the whole chain of the calls

e5l (approved):
LGTM in general. We can improve samples a bit

codeSnippets/snippets/client-call-id/src/main/kotlin/com/example/Application.kt
vnikolova: Do you mean like this:

```kotlin
fun main() {
        getRequestWithCustomCallId()
        //...
}

fun getRequestWithCustomCallId() {
    runBlocking {
        val client = HttpClient(CIO) {
            install(CallId) {
                generate { "call-id-client" }
                addToHeader(HttpHeaders.XRequestId)
                intercept { request, callId -> request.parameter("callId", callId) }
            }
        }
        val response: HttpResponse = client.get("http://0.0.0.0:8090/call")
        println(response.bodyAsText())

        val responseFromDoubleCall: HttpResponse = client.get("http://0.0.0.0:8090/double-call")
        println(responseFromDoubleCall.bodyAsText())
        client.use {
            client.close()
        }
    }
}
//...
```

e5l: fun Closeable.use(block: () -> Unit) {
}

e5l (approved):

PR #353, opened 2023-10-16
Refactor names to comply with Kotlin naming conventions

artemmukhin:
While checking out the Ktor documentation, I noticed some variable names that didn't quite follow Kotlin's naming conventions. So, I went through the codebase and gave them a little update.

Feedback is appreciated!


vnikolova:
Thank you for the contribution, @artemmukhin. While it may be that the names don't follow Kotlin's naming conventions, they are generated by the Ktor plugin. So, refactoring them in the docs will cause inconsistency with the generated code. If you think the naming should be refactored for the Ktor plugin generator, please feel free to create an issue on [YouTrack](https://youtrack.jetbrains.com/issues/KTOR).

PR #355, opened 2023-10-16
There is no SOCKS proxy support on Darwin.

idf3d:
See [ProxySupportCommon.kt#L21](https://github.com/ktorio/ktor/blob/491acd80bda5196a93b418d111ccda5da347fb0a/ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/ProxySupportCommon.kt#L21), SOCKS not supported by Darwin.


vnikolova:
Thank you for your contribution @idf3d!

PR #379, opened 2023-11-13
Update Ktor dependencies where the Ktor Gradle plugin is applied

vnikolova:
The BOM dependency is already described [here](https://ktor.io/docs/server-dependencies.html#dependencies-consistency).

- Removed `ktor_version` and updated the dependency names where the Ktor Gradle plugin is applied.


e5l (approved):
lgtm

PR #380, opened 2023-11-14
KTOR-6459 Update Kotlin version to 1.9.20

vnikolova:
- update kotlin version in version files
- update plugin versions
- remove deprecated `kotlin-js` dependency in `client-engine-js`


codeSnippets/snippets/tutorial-client-kmm/gradle/libs.versions.toml
```diff
@@ -1,6 +1,6 @@
 [versions]
 agp = "8.1.1"
-kotlin = "1.9.10"
+kotlin = "1.9.20"
 kotlinx = "1.7.3"
```
e5l, line 4: just noted, it looks like it should be `coroutines` instead of `kotlinx`

e5l (approved):
lgtm

PR #395, opened 2023-11-29
Update DatabaseSingleton.kt to use companion object

maxmccord:
Found this issue while following the persistence tutorial: https://ktor.io/docs/interactive-website-add-persistence.html#connect_db

As previously written, the code does not compile. Static functions should be placed within a companion object.


maxmccord:
I'm sorry, my mistake - in my local code I'd defined DatabaseSingleton as a `class` rather than an `object`. I'll close this request.

PR #405, opened 2024-01-12
fix(jwt): Added -jvm postfix for Maven dependencies

sampie777:
Added the missing `-jvm` postfix for the dependencies used by Maven. 
Without this postfix, Maven cannot compile the project, as all classes etc. are missing from the default `ktor-server-auth-jwt` artifact.


e5l (approved):

e5l:
Hey @sampie777, thanks for the catch! LGTM

PR #407, opened 2024-01-18
KTOR-6520 Get started with Ktor Server tutorial

vnikolova:
[KTOR-6520](https://youtrack.jetbrains.com/issue/KTOR-6520/Documentation-Restructuring-Getting-started-with-Ktor-Server)

[Preview on Staging (Updated)](https://ktor.shared.aws.intellij.net/docs/server-create-a-new-project.html)

Added a new topic with the "Get started with Ktor Server" tutorial:

- including screenshots for both light and dark theme
- a new project example has been added
- the list of tutorials represents what is currently available and will be adjusted once the rest of the tutorials are added
- the [Next Steps](https://writerside.labs.jb.gg/staging/staging/ktor/2.3.7/server-get-started.html#next-steps) section currently points to a link to "Creating HTTP APIs", which will be replaced with "Understanding Routing and Requests"

Edit:

- Added a collapsible chapter for creating a new project with intelliJ IDEA Ultimate. Creating a project with the web generator is expanded by default.
- The title of the tutorial has been renamed to "Create, open and run a new Ktor project".
- The tutorial is placed under an empty collapsible nav item "Getting started with Ktor Server".


garthgilmourni (changes requested):
This looks great! Thank you very much.

I can see three things that might need changed:
1. When this content was added, as the final page of the Ktor Project Generator, it was decided to remove stages 2 and 6 in the [Unpack and run the project](https://writerside.labs.jb.gg/staging/staging/ktor/2.3.7/server-get-started.html#unpacking) section. This was done to shorten the sequence and simplify the document. For consistency I think we should do the same here.
2. I think stage 5 in the above section (using the chmod command) is only required on MacOS / UNIX and not Windows.
3. There is a reference to the _Common Problems document_ - which we never got around to writing.

e5l (changes requested):
Looks very nice! Thank you for the PR. I noticed a little problem with navigation:
image
The page is on the collapsable item, and it's not clear why I should click it. For other entries (like `Ktor Server`) we usually redirect to the first nested page. Could we make a separate entry for the tutorial?

e5l:
I think we also should move `Creating a new Ktor Project` tutorial and clarify the structure to highlight IDE Wizard and start.ktor.io website

vnikolova:
@e5l Thank you for the comments. You're right, at the moment it is a bit confusing having the topic at the collapsible item. What do you think of having only the introduction there including the list of tutorials? It is [recommended by Writerside](https://www.jetbrains.com/help/writerside/table-of-contents.html#wrapper-topic) that grouped topics have an introduction (no empty wrapper topics).

Regarding the `Creating a new Ktor Project` topic: I moved it outside of the group because it generates a project with a different name and plugins. If it is part of the group, should it be preceding the first tutorial from the guide?

meilalina:
Hi, @vnikolova! Thank you very much for updating the tutorial, it looks great! 

Leonid, nice catch on the collapsible item. It sparked a lot of thoughts, as I hadn't gone through the structure update before. So, apologize for the long comment, but I'll break it down into a few separate points for easier discussion.

1. About the "Creating a new Ktor project" guide - having it separate might confuse users since it's similar to "Getting Started...". Too many choices might lower conversions. People might skip the detailed "Getting Started" tutorial for the shorter "Create new project" one, specifically as it kinda duplicates the beginning of the tutorial. What if we merge them? "Creating a new project" could be the first article in "Getting Started," especially now that we're dropping it from the collapsible item. Basically, my suggestion is to keep the current structure.

2. In the old guide, IntelliJ IDEA Ultimate seemed to be preferred for project creation, but the new one favors the web generator. Why the switch? Should we still highlight Ultimate, or is there a reason to shift focus?

3. For now, I wouldn't add anything to the collapsible item. It might make our tutorial inconsistent with the rest of the docs, where we use empty wrappers. Fixing all of them seems like a huge task, and probably should be addressed separately.

4. As a side note, we could later rethink our overall document structure to experiment and see if it boosts engagement/conversions. Maybe a simple, top-level "Getting Started" tutorial, like on kotlinlang.org, could be a thing to try. We can discuss this later, as a part of other website update experiments.

vnikolova:
@e5l @meilalina What do you think of renaming the current "Create a new Ktor project" topic to "Installation"? So "Installation" would include 2 sections of how to generate a new project with intelliJ IDEA Ultimate and the online generator. I think it is important to have extended information on how to add plugins, which is missing from the "Get started" tutorial.

So if we have "Installation" as a separate topic, and "Getting started with Ktor Server"  as an empty collapsible item, then the first tutorial can be renamed to "Create, open and run a new Ktor project".

meilalina:
@vnikolova I need some time to think about the 'Installation' suggestion. I'll get back to you with an answer.

Meanwhile, I've received suggestions for SEO improvements and would appreciate your help in implementing them. The current issue is that our 'Getting Started' article isn’t appearing in the top search results for branded queries (e.g., when someone searches for Ktor). This is not great as it likely leads to missed opportunities for converting searchers into users (or at least tries). 

Here is what we can do to improve the situation:
1. Make sure the meta description is clear and relevant. We can include "how to start/get started/how to create ktor project / etc." @vnikolova, could you help with it or should I work on it with Ekaterina Zaikina?
2. Add "Ktor" to H2, e.g. “Unpack and run your Ktor project”, “Open, explore and run the Ktor project in IntelliJ IDEA”.
3. Add backlinks to the Getting Started from our blog posts, for example:  
https://blog.jetbrains.com/ktor/2022/04/11/ktor-2-0-released/
https://blog.jetbrains.com/ktor/2022/08/12/ktor-2-1-0-released-and-it-comes-with-goodies/
https://blog.jetbrains.com/ktor/2022/12/16/ktor-2023-roadmap/
I'll handle the backlinks myself. Backlinks also do not stop us from publishing. 
4. Add backlinks from other articles in our Docs. @vnikolova since you might be more familiar with our documentation, perhaps you have some ideas for where these could be effectively placed. I'll also review articles and share my thoughts. 
5. @vnikolova are we going to keep the version in the URL ktor/2.3.7/server-get-started.html after the release? That would be better to remove it.

vnikolova:
@meilalina 
- Regarding the meta description - I can certainly add it. Do you mean that you want to include these words in the description or also as keywords?
- Yes, I have some ideas. I will go through the docs and take a note where we can reference it.
- The version won't be in the URL. I will deploy my next changes to the main staging environment so you can see exactly how it would look. The idea is that the tutorial will be accessible for both versions `2.3.7` and `3.0.0-beta-1`.

e5l:
@vnikolova "Installation" is usually not used with libraries. I would try something like "Setting up a project." 

It also feels like all tutorials are getting started. Could you tell me how we can try fitting all the tutorials there? Maybe we can merge it with the difference under collapsable blocks. Otherwise, the starting point is not clear.

meilalina:
Hi @vnikolova 

I did a quick research and it appears that people don't search for _ktor install/installation_ a lot. The query _ktor create project_ is somewhat more popular, but it's still a very infrequent request.

I've also reviewed the tutorials of other frameworks (Spring Boot, Quarkus, http4k), and those that mention installation typically include it as a part of their Getting Started guides.

Therefore, I don't think we should separate the article from the Getting Started tutorial. Let's see if they can be merged. I like Leonid's suggestion of using collapsible blocks.

meilalina:
Hi @vnikolova ! Thanks for the update. I wouldn't hide the generator path under the collapsible item, and only hide the IDEA path. Is that possible?

vnikolova:
> Hi @vnikolova ! Thanks for the update. I wouldn't hide the generator path under the collapsible item, and only hide the IDEA path. Is that possible?

@meilalina Yes, certainly. You can already see the change reflected on the staging environment.

meilalina:
Thank you @vnikolova ! Looks good to me

e5l (approved):
looks nice

garthgilmourni (approved):
Looks good to me.

PR #423, opened 2024-02-09
Add the meaning of the Ktor name in FAQ

vnikolova:
I just saw the question asked again in Slack so decided to add it to the FAQ.

Also:
- Replaced the pronunciation to be represented phonetically, as Kay-tor doesn't make it obvious the [a] is pronounced as [e]
- Formatted the file to fix autodetected issues


e5l (approved):
Nice catch! LGTM

PR #443, opened 2024-03-14
Fix 3.0.0-beta-1 migration guide link

twyatt:
Link contained an erroneous greater than sign at the end of the URL.


vnikolova (approved):
Thank you for the fix @twyatt!

PR #450, opened 2024-03-16
Fix broken API links in the "Choosing an engine" topic

oak-cassia:
- Choosing an engine
  - [embeddedServer﻿](https://ktor.io/docs/engines.html#embeddedServer)
  - [ApplicationEngine.Configuration](https://ktor.io/docs/engines.html#configure-engine)


vnikolova (approved):
Well spotted! Thank you for your contribution @oak-cassia

PR #481, opened 2024-06-06
Refactor/update ktor server proguard simple

ellet0:
Update the Ktor Server Proguard example:

1. Avoid using `ignorewarnings` and `dontwarn`, instead fix the warnings by configuring the Kotlin standard library and the required Java modules in Proguard configurations, this will not only fix warnings but avoid many bugs and issues that are hard to debug as the message and errors can become quite unclear because Proguard doesn't have access to the used Java modules that are part of the JDK, for more details, see this [comment](https://github.com/square/okio/issues/1298#issuecomment-1652182091)
2. Include the official rules of Kotlinx serialization (https://github.com/Kotlin/kotlinx.serialization/blob/master/rules/common.pro) and coroutines (https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/jvm/resources/META-INF/proguard/coroutines.pro)
3. Update the Proguard rules to get the example working without using `ignorewarnings` to fix many bugs. This new version of Proguard rules can still be improved further. The rules are configured to build the project without any issues for the following libraries:
    ```kotlin
    implementation("io.ktor:ktor-server-auth-jvm")
        implementation("io.ktor:ktor-server-core-jvm")
        implementation("io.ktor:ktor-server-webjars-jvm")
        implementation("org.webjars:jquery:3.2.1")
        implementation("io.github.smiley4:ktor-swagger-ui:2.9.0")
        implementation("io.ktor:ktor-server-resources-jvm")
        implementation("io.ktor:ktor-server-host-common-jvm")
        implementation("io.ktor:ktor-server-status-pages-jvm")
        implementation("io.ktor:ktor-server-cors-jvm")
        implementation("io.ktor:ktor-server-caching-headers-jvm")
        implementation("io.ktor:ktor-server-compression-jvm")
        implementation("io.ktor:ktor-server-call-logging-jvm")
        implementation("io.ktor:ktor-server-content-negotiation-jvm")
        implementation("io.ktor:ktor-serialization-kotlinx-json-jvm")
        implementation("io.ktor:ktor-server-websockets-jvm")
        implementation("io.ktor:ktor-server-auth-ldap-jvm")
        implementation("io.ktor:ktor-server-netty-jvm")
        implementation("ch.qos.logback:logback-classic:$logback_version")
    ```
    I haven't tested all the functionality for each one though, fixing the warnings can catch some common bugs though.
5. Update Proguard to `7.5.0` to support Java 22, We could update the Gradle wrapper, Kotlin, also the versions are hardcoded for all used tools like Ktor Gradle Plugin, I suggest using Gradle version catalogs `gradle/libs.versions.toml`
6. Use the official Proguard [https://github.com/Guardsquare/proguard/blob/master/examples/application-kotlin/proguard.pro](application-kotlin) example as a reference
7. Create a new task called `runMinimizedJar` and replace `minimizedJar` with `buildMinimizedJar` to maintain consistency with `buildFatJar` and `runFatJar`
8. Avoid hardcoding the name of the minimized JAR file
9. Tested with Java 11, 17, 21, and Kotlin 2.0.0, Gradle 8.8, and tested with the version tool versions in this code snippet.
10. Group both `buildMinimizedJar` and `runMinimizedJar` in `ktor`:
image
to be along with `buildFatJar` and `runFatJar`

I suggest renaming the `proguard` module to `ktor-server-proguard` or `server-proguard` or something similar for clarity as this is for the Ktor server and won't work directly for Ktor Client, notice that you will need to include more rules depending on the engine and the libraries, for example If you're using OkHttp as the engine in Ktor client, you will need to include the rules from (https://square.github.io/okhttp/features/r8_proguard/) and Okio, as well as OkHttp, depends on it.

**Not related**:
I noticed Ktor somehow depends on Java Swing/Desktop, all imports from `org.mozilla.javascript.tools.debugger` use it, while this may sound not an issue, it could cause some issues, like making the bundle size bigger when using `jlink` to include the required Java modules only in the application instead of fully installing Java on the machine, an easy solution is to split this functionality into a different module and make it optional to use.

While this example is referenced in Ktor docs, it's also applicable in the [ktor-samples](https://github.com/ktorio/ktor-samples) repository.

Feel free to share your suggestions or request any changes.

### DIFF
--- a/codeSnippets/snippets/proguard/README.md
+++ b/codeSnippets/snippets/proguard/README.md
@@ -22,11 +22,15 @@ this example:
 3. [Proguard](https://github.com/Guardsquare/proguard) Kotlin Example: https://github.com/Guardsquare/proguard/blob/master/examples/application-kotlin/proguard.pro
 4. [Google Guava](https://github.com/google/guava): https://github.com/google/guava/wiki/UsingProGuardWithGuava
 
-This example is configured to work with Kotlinx coroutines and serialization, required modules that are part of the JDK
+This example is already configured to work the libraries mentioned above, required modules that are part of the JDK
 and some libraries that Ktor server depends on.
 
+> If you're using Ktor Client with OkHttp as an engine in Ktor Server, you need to include
+> [OkHttp Proguard](https://square.github.io/okhttp/features/r8_proguard/) Rules, otherwise you will get warnings that
+> cause the proguard task build to fail.
+
 If you're using additional libraries that use features such as 
-reflections or dynamic class loading, You will get a warning that needs to be solved, some common ways to solve this:
+reflections or dynamic class loading, You will usually get warnings that need to be solved, some common ways to solve this:
 
 1. You can see if the library supports Proguard, take a look at the docs, resources of the module you want to include.
 `README.md` file, or some common folders such as `proguard` or `rules`, also in `META-INF/proguard` in the JAR file, which
--- a/codeSnippets/snippets/proguard/README.md
+++ b/codeSnippets/snippets/proguard/README.md
@@ -3,12 +3,30 @@
 A sample project packed as a JAR using the [Ktor Gradle plugin](https://ktor.io/docs/fatjar.html) and minimized using [ProGuard](https://www.guardsquare.com/manual/home).
 > This sample is a part of the [codeSnippets](../../README.md) Gradle project.
 
+This example is configured to work with Kotlinx coroutines and serialization, required modules that are part of the JDK
+and some libraries that Ktor server depends on.
+
+if you're using additional libraries that use features such as 
+reflections or dynamic class loading, You will get a warning that needs to be solved.
+
+For this example, the Fat JAR size is (15.2 MB) and the minimized JAR (9 MB). This will
+be different depending on your project, dependencies, and Proguard configurations.
+
 ## Running
 
 To build a minimized JAR and run the application, execute the following commands:
+
 ```bash
-./gradlew :proguard:minimizedJar
+./gradlew :proguard:buildMinimizedJar
 java -jar snippets/proguard/build/libs/my-application.min.jar
 ```
 
+or run the JAR directly using a Gradle task:
+
+```bash
+./gradlew :proguard:runMinimizedJar
+```
+
 Then, navigate to [http://localhost:8080/](http://localhost:8080/) to see the sample home page.
+
+> Replace `gradlew` with `gradlew.bat` if you're using **Microsoft Windows**
--- a/codeSnippets/snippets/proguard/README.md
+++ b/codeSnippets/snippets/proguard/README.md
@@ -3,14 +3,52 @@
 A sample project packed as a JAR using the [Ktor Gradle plugin](https://ktor.io/docs/fatjar.html) and minimized using [ProGuard](https://www.guardsquare.com/manual/home).
 > This sample is a part of the [codeSnippets](../../README.md) Gradle project.
 
+**Proguard** is a tool to shrink, obfuscate, and optimizer for Java bytecode and Android apps, it's the default solution
+used in frameworks and projects
+like [Compose Desktop](https://github.com/JetBrains/compose-multiplatform/blob/master/tutorials/Native_distributions_and_local_execution/README.md#minification--obfuscation).
+
+Proguard will need rules to keep some classes that use **reflection** and **dynamic access
+(dynamic class loading)** to work correctly.
+
+Proguard will give you **warnings** that need to be **either solved or suppressed**. 
+This is usually caused by missing class references, Java module that's used yet not included, or when it detects usages
+that can't be automated.
+
+Libraries and frameworks might have rules that need to be included to work properly, the following rules are included in
+this example:
+
+1. [Kotlinx Coroutines](https://github.com/Kotlin/kotlinx.coroutines): https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/jvm/resources/META-INF/proguard/coroutines.pro
+2. [Kotlinx Serialization](https://github.com/Kotlin/kotlinx.serialization): https://github.com/Kotlin/kotlinx.serialization/blob/master/rules/common.pro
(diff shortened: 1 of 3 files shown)


e5l (approved):
That's really nice revamp all proguard configuration! Thank you for the PR!


LGTM

ellet0:
> That's really nice revamp all proguard configuration! Thank you for the PR!
> 
> 
> LGTM

I believe it can be improved further before merging, for example:

1. We could use `Paths` and `JavaVersion` to achieve some functionalities.
2. Instead of copy and paste the rules of Kotlinx coroutines and serialization manually, it will be already included in `META-INF/proguard` in the Jar regarding if we're using Proguard or not, we could use them so in case we update the libraries, the rules will also updated if there is any changes and will be included.

Let me know if you're interested in those changes to get them into the example so I update the PR.

e5l:
I agree. It would be great if you could log an issue for that

codeSnippets/snippets/proguard/README.md
```diff
@@ -3,12 +3,30 @@
 A sample project packed as a JAR using the [Ktor Gradle plugin](https://ktor.io/docs/fatjar.html) and minimized using [ProGuard](https://www.guardsquare.com/manual/home).
 > This sample is a part of the [codeSnippets](../../README.md) Gradle project.
 
+This example is configured to work with Kotlinx coroutines and serialization, required modules that are part of the JDK
+and some libraries that Ktor server depends on.
+
+if you're using additional libraries that use features such as 
+reflections or dynamic class loading, You will get a warning that needs to be solved.
```
vnikolova, line 10: Here do you mean that the user would need to include more rules to fix the warning?

vnikolova (changes requested):
Thank you for the contribution and elaborate explanation @ellet0 ! 

It might be useful to include the links to the official rules of Kotlinx serialization and coroutines in the README file, as well  as the list of libraries you tested this against and this paragraph:
"You will need to include more rules depending on the engine and the libraries. For example, If you're using OkHttp as the engine in Ktor client, you will need to include the rules from (https://square.github.io/okhttp/features/r8_proguard/) and Okio as well as OkHttp depends on it." 
Can you add these?

And I agree with your suggestion to rename the example to `server-proguard` and using Gradle version catalogs. We will address these changes in a separate PR. 👍

codeSnippets/snippets/proguard/README.md
ellet0: > Here do you mean that the user would need to include more rules to fix the warning?

The developer probably use more libraries that are not part of the example, he or she might not familiar with Proguard as it's known as Android tool,including this note to indicate that the warnings as Proguard won't be able to solve any usages that use dynamic access (dynamic class loading) or reflections and need to be solved.

Proguard will be able to detect most common usages, see (https://www.guardsquare.com/manual/home) for more details.

The example is shown in the documentation of the Ktor server, so it's why I included details that's related to Proguard.

Let me know if I should remove this note.

vnikolova: Thanks for clarifying. You can keep the note and add a reference to the Proguard docs for more information.

ellet0:
> I agree. It would be great if you could log an issue for that, so it will not be lost

Solved.

ellet0:
Regarding the note of using `JavaVersion` instead of `System.getProperty("java.version")` to check in a type-safe way that's more readable, I didn't update the code as Progaurd needs access to the JDK path that's used by Gradle instead of the one in system environment path.

I'm not sure if there is an API in Gradle to get the Path of the used JDK, the developer might want to use a Java version for a project that's not the same as the one configured to be used globally, using something like `JavaVersion.current().isCompatibleWith(JavaVersion.VERSION_1_9)` will check the java version that's used by Gradle to run Gradle daemon or the task.

Which use different versions between the two and might cause an unexpected bug in some rare cases where the developer includes a module that's available in a certain Java version and not available in the project Java Version, instead I used the following:

```kotlin
val javaHome = System.getProperty("java.home")
    val javaVersion = System.getProperty("java.version")
    if (javaVersion.startsWith("1.")) {
// A version before Java 9
}
```

We also could include all available Java modules in the current JDK instead of explicitly including the required or used modules.

vnikolova (approved):

vnikolova:
Thank you for the comments @ellet0! Your PR is now merged 🎉

ellet0:
> Thank you for the comments @ellet0! Your PR is now merged 🎉

I appreciate your consideration. It's been a rewarding experience.

Should we use Proguard rules from the JAR file in `META-INF/proguard` instead of copy and paste them manually in `proguard.pro`?

I already did this with the following code snippet:

```kotlin
JarFile(fatJarFile.get().asFile).use { jarFile ->
                val generatedRulesFiles =
                    jarFile
                        .entries()
                        .asSequence()
                        .filter { it.name.startsWith("META-INF/proguard") && !it.isDirectory }
                        .map { entry ->
                            jarFile.getInputStream(entry).bufferedReader().use { reader ->
                                Pair(reader.readText(), entry)
                            }
                        }.toList()
                
                val buildProguardDirectory =
                    layout.buildDirectory
                        .dir("proguard")
                        .get()
                        .asFile
                if (!buildProguardDirectory.exists()) {
                    buildProguardDirectory.mkdir()
                }
                generatedRulesFiles.forEach { (rulesContent, rulesFileEntry) ->
                    val rulesFileNameWithExtension = rulesFileEntry.name.substringAfterLast("/")
                    val generatedProguardFile = File(buildProguardDirectory, "generated-$rulesFileNameWithExtension")
                    if (!generatedProguardFile.exists()) {
                        generatedProguardFile.createNewFile()
                    }
                    generatedProguardFile.bufferedWriter().use { bufferedWriter ->
                        bufferedWriter.appendLine("# Generated file from ($rulesFileEntry) - manual changes will be overwritten")
                        bufferedWriter.appendLine()

                        bufferedWriter.appendLine(rulesContent)
                    }

                    configuration(generatedProguardFile)
                }
            }
```

This will help make the rules for Kotlinx coroutines, serialization, and others up to date by only updating the libraries without changing `proguard.pro` file.

Regarding separating `proguard` example into `server-proguard`, we also could have a folder called `proguard` with two folders in it `client` and `server` because Ktor client can be used with Proguard too.

PR #484, opened 2024-06-15
KTOR-6817 Explain why the `auth-form` authentication is referred on t…

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-6817

### DIFF
--- a/topics/server-session-auth.md
+++ b/topics/server-session-auth.md
@@ -90,7 +90,7 @@ The `session` authentication provider exposes its settings via the [SessionAuthe
 ### Step 4: Save user data in a session {id="save-session"}
 
 To save information about a logged-in user to a session, use the [call.sessions.set](server-sessions.md#use_sessions) function.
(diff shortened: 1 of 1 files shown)


topics/server-session-auth.md
```diff
@@ -90,6 +90,8 @@ The `session` authentication provider exposes its settings via the [SessionAuthe
 ### Step 4: Save user data in a session {id="save-session"}
 
 To save information about a logged-in user to a session, use the [call.sessions.set](server-sessions.md#use_sessions) function.
+> The definition of `auth-form` authentication isn't detailed on this page; it is provided here solely for illustrative purposes.
```
vnikolova, line 93: Instead of the first sentence, perhaps we can note that "this example shows a simple authentication flow using a form.". Then from the next sentence is clear where to find more details.

vnikolova (commented):
Thanks for the PR 🫶

vnikolova (approved):

PR #487, opened 2024-06-27
Chore/proguard server simple minor change

EchoEllet:
While working on a side project that has similar changes sent in #481, noticed that I'm using `java.nio.file.Paths` to build the path as String and `java.io.File` to check if the file exists which is a bit inconsistent, sending this change here as well, this PR has two minor changes

1. Avoid converting the path of the `Path` class to a `String` to create the `File`, instead use the `Files` class directly to check if the file exists, throw `java.nio.file.NoSuchFileException` as an alternative of `java.io.FileNotFoundException`.
While this code doesn't import `java.io.File` as it still uses `org.gradle.api.file.RegularFile` which uses the `File` class. We could also use `file` function from Gradle which return `File`
2. If the developer removed the line `dontobfuscate()`, Progaurd will generate a map file using `printmapping` function that can be used with [Proguard Retrace](https://www.guardsquare.com/manual/tools/retrace) to get a readable stack trace that makes the debugging process easier for development, previously the file will be `my-application.map`, now `my-application.min.map` since this map file is specific to the Proguard minimized JAR file

**Side note**:
If you want a readable stack trace with the source file and line number, in addition to using **Proguard Retrace**, you need to include the following in `proguard.pro`:

```
-keepattributes SourceFile,LineNumberTable
```

This will increase the Minimized JAR size a bit and you still need to use Proguard Retrace.

We could provide a link to Proguard Retrace in the README.md of this simple.

### DIFF
--- a/codeSnippets/snippets/proguard/build.gradle.kts
+++ b/codeSnippets/snippets/proguard/build.gradle.kts
@@ -1,4 +1,5 @@
 import java.io.FileNotFoundException
+import java.nio.file.Files
 import java.nio.file.Paths
 
 val ktor_version: String by project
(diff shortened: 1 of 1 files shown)


vnikolova (approved):
Looks neat 👌  Thanks for the contribution @EchoEllet !

EchoEllet:
> Looks neat 👌 Thanks for the contribution @EchoEllet !

Thank you for the review and for maintaining the project.

PR #491, opened 2024-07-02
fix typo in configure database section of ktor server docs

beradeep:
fix typo `Databse.connect()` to `Database.connect()`


vnikolova (approved):
Thanks for the fix @beradeep 🙏

PR #500, opened 2024-07-26
KTOR-7239 Add tutorial "First steps with Kotlin RPC"

vnikolova:
[KTOR-7239](https://youtrack.jetbrains.com/issue/KTOR-7239)

[Preview on Staging](https://ktor.shared.aws.intellij.net/docs/tutorial-first-steps-with-kotlin-rpc.html)

- modified article content to include procedures
- added project examples
- added screenshots for light and dark theme
- added meta, card and link descriptions
- added new topic to the [Welcome page](https://ktor.shared.aws.intellij.net/docs/welcome.html)


e5l (approved):

codeSnippets/snippets/tutorial-kotlin-rpc-client/src/main/kotlin/Main.kt
```diff
@@ -0,0 +1,60 @@
...
+        rpcConfig {
+            serialization {
+                json()
+            }
+        }
+    }
+
+    val pizzaShop: PizzaShop = client.withService<PizzaShop>()
+
+/*
```
Stexxe, line 36: Commented code

codeSnippets/snippets/tutorial-kotlin-rpc-client/src/main/kotlin/model/PizzaShop.kt
```diff
@@ -0,0 +1,19 @@
+package com.example.model
+
+import kotlinx.coroutines.flow.Flow
+import kotlinx.serialization.Serializable
+import kotlinx.rpc.RPC
+
+interface PizzaShop : RPC {
+    /*
+    suspend fun orderPizza(pizza: Pizza): Receipt
```
Stexxe, line 9: Commented code

codeSnippets/snippets/tutorial-kotlin-rpc-server/src/main/kotlin/Main.kt
```diff
@@ -0,0 +1,72 @@
...
+import kotlinx.rpc.serialization.json
+import kotlinx.rpc.transport.ktor.server.RPC
+import kotlinx.rpc.transport.ktor.server.rpc
+import kotlin.coroutines.CoroutineContext
+
+class PizzaShopImpl(
+    override val coroutineContext: CoroutineContext
+) : PizzaShop {
+    /*
+    override suspend fun orderPizza(pizza: Pizza): Receipt {
```
Stexxe, line 22: Commented code

codeSnippets/snippets/tutorial-kotlin-rpc-server/src/main/kotlin/Main.kt
```diff
@@ -0,0 +1,72 @@
...
+            }
+        }
+        return flow {}
+    }
+}
+
+fun main() {
+    embeddedServer(Netty, port = 8080) {
+        module()
+        println("Server running")
```
Stexxe, line 54: println

codeSnippets/snippets/tutorial-kotlin-rpc-server/src/main/kotlin/model/PizzaShop.kt
```diff
@@ -0,0 +1,17 @@
+import kotlinx.coroutines.flow.Flow
+import kotlinx.rpc.RPC
+import kotlinx.serialization.Serializable
+
+interface PizzaShop : RPC {
+    /*
+    suspend fun orderPizza(pizza: Pizza): Receipt
```
Stexxe, line 7: Commented code

Stexxe (approved):

topics/tutorial-first-steps-with-kotlin-rpc.topic
```diff
@@ -0,0 +1,600 @@
...
+        </p>
+    </tldr>
+    <p>
+        Kotlin RPC (Remote Procedure Call) is a new and exciting addition to the Kotlin ecosystem, which builds on well-established foundations
+        and runs on the <code>kotlinx.rpc</code> library.
+    </p>
+    <p>
+        The <code>kotlinx.rpc</code> library enables you to make procedure calls over network
+        boundaries using nothing but regular Kotlin language constructs. As such, it provides an alternative to both
+        REST and Google RPC (gPRC) while (in upcoming functionality) offering opportunities to interoperate with both.
```
Mr3zee, line 42: The wording is a bit misleading I think. It is not obvious from the first reading, that both gRPC and REST integrations, are not supported yet.

topics/tutorial-first-steps-with-kotlin-rpc.topic
```diff
@@ -0,0 +1,600 @@
...
+                <li>
+                    A means to specify the types used for parameters and return values. Once again, you could use a
+                    language-neutral standard. However, it may be simpler to annotate standard data type declarations in
+                    the current language.
+                </li>
+                <li>
+                    Helper classes that will be used to convert the procedure invocation into a format that can be sent
+                    over the network and unpack the return value that results. These are often referred to as
+                    <format style="italic">client stubs</format>
+                    . They may be created at build time or dynamically at runtime.
```
Mr3zee, line 102: I think we can make this a bit clearer. From the first read, it is not obvious, who is the stub - the serialized procedure or the one who serializes. While the latter is true, it may not be so understandable for the first-time reader, who is not well into the RPC business

topics/tutorial-first-steps-with-kotlin-rpc.topic
```diff
@@ -0,0 +1,600 @@
...
+        </chapter>
+        <chapter title="Add the shared types" id="shared-types">
+            <p>
+                The heart of any RPC project is the interface that defines the procedures to be called remotely, along
+                with the types used in the definition of those procedures.
+            </p>
+            <p>
+                These types will need to be placed in both the client and server projects. Ideally, you want a build
+                process that packages them into a separate library, which can be shared as needed. However, for
+                simplicity of this example, you can copy and paste.
```
Mr3zee, line 189: > However, for simplicity of this example, you can copy and paste.

I would disagree, that it is simpler. While it may be true from the build setup point of view, the two other aspects are hurt:
- This approach hides the biggest advantage of the Kotlin RPC - write once, use everywhere, 
- It instead uses copy-paste thing, which is generally considered bad and error-prone. I don't we should endorse such an approach even in a toy example project.

I see why it is easier in the example. Still, I think that it would hurt the reader more in the long run. Besides, we already have two modules, adding one more it not that big of a deal, isn't it?

topics/tutorial-first-steps-with-kotlin-rpc.topic
```diff
@@ -0,0 +1,600 @@
...
+                    <code-block lang="kotlin"
+                                src="snippets/tutorial-kotlin-rpc-client/src/main/kotlin/model/PizzaShop.kt"
+                                include-lines="1-7,9,13-19"
+                    />
+                    <p>
+                        The interface needs to extend the <code>RPC</code> type from the <code>kotlinx.rpc</code> library. The
+                        reason for this will be explained later.
+                    </p>
+                    <p>
+                        Assuming you are using <a href="https://github.com/Kotlin/kotlinx.serialization"><code>kotlinx.serialization</code></a>
```
Mr3zee, line 214: No other serialization is supported, and we have no immediate plans for this

codeSnippets/snippets/tutorial-kotlin-rpc-client/src/main/kotlin/Main.kt
```diff
@@ -0,0 +1,60 @@
...
+import kotlinx.rpc.client.withService
+import kotlinx.rpc.serialization.json
+import kotlinx.rpc.transport.ktor.client.rpc
+import kotlinx.rpc.transport.ktor.client.rpcConfig
+import kotlinx.rpc.streamScoped
+
+fun main() = runBlocking {
+    val ktorClient = HttpClient {
+        install(WebSockets)
+    }
```
Mr3zee, line 18: We have `installRPC` function instead, see https://kotlin.github.io/kotlinx-rpc/transport.html#client

topics/tutorial-first-steps-with-kotlin-rpc.topic
```diff
@@ -0,0 +1,600 @@
...
+                matching plugin is loaded:
+            </p>
+            <code-block lang="kotlin"
+                        src="snippets/tutorial-kotlin-rpc-client/src/main/kotlin/Main.kt"
+                        include-lines="16-18"
+            />
+            <p>
+                Having created this Ktor client, you then create an <code>RPCClient</code> object for invoking remote
+                procedures. You will need to configure the location of the server and the mechanism being used to
+                transfer information:
```
Mr3zee, line 266: From the code example, it is not obvious, which part is an `RPCClient`. I think it would be beneficial to add the type explicitly in the code

topics/tutorial-first-steps-with-kotlin-rpc.topic
```diff
@@ -0,0 +1,600 @@
...
+                            no use of reflection.
+                        </li>
+                        <li>
+                            The Ktor plugin is used to build fat JARs that bundle the app with all its dependencies.
+                        </li>
+                        <li>
+                            KSP is the Kotlin Symbol Processing API used to develop the <code>kotlinx.rpc</code> plugin.
+                        </li>
+                        <li>
+                            The RPC plugin is needed to create the client- and server-side stubs.
```
Mr3zee, line 403: We don't have server-side stubs, only client-side

topics/tutorial-first-steps-with-kotlin-rpc.topic
```diff
@@ -0,0 +1,600 @@
...
+                REST APIs are built on a series of conventions that make them relatively easy to create, document,
+                monitor, debug, and test. There is a huge ecosystem of tools to support this.
+            </li>
+        </list>
+        <p>
+            These trade-offs mean that Kotlin RPC is best used in two scenarios. Firstly, in KMP clients using <a
+                href="https://www.jetbrains.com/lp/compose-multiplatform/">Compose Multiplatform</a>, and secondly,
+            amongst collaborating microservices in the cloud. Future developments in <a
+                href="https://kotlinlang.org/docs/wasm-overview.html">Kotlin/Wasm</a> may make kotlin.rpc more
+            applicable to browser-based applications.
```
Mr3zee, line 505: kotlinx.rpc

topics/tutorial-first-steps-with-kotlin-rpc.topic
```diff
@@ -0,0 +1,600 @@
...
+            Buffers (protobuf) is used to define data structures and message payloads using a language-neutral Interface
+            Definition Language (IDL). These IDL definitions can be converted into a wide variety of programming
+            languages and are serialized using a compact and efficient binary format. Microservice frameworks like
+            Quarkus and Micronaut already have support for gRPC.
+        </p>
+        <p>
+            It would be difficult for Kotlin RPC to compete with gRPC, and there would be no benefit in this for the
+            Kotlin community. Thankfully, there are no plans to do this. Instead, the intention is for kotlinx.rpc to be
+            compatible and interoperable with gRPC. It will be possible for kotlinx.rpc services to use gRPC as their
+            networking protocol and for kotlinx.rpc clients to call gRPC services. kotlin.rpc will use <a
```
Mr3zee, line 520: kotlinx.rpc

topics/tutorial-first-steps-with-kotlin-rpc.topic
```diff
@@ -0,0 +1,600 @@
...
+                On the server side, navigate to the
+                <path>Main.kt</path>
+                file and implement this functionality by storing the current orders in a map of lists:
+                <code-block lang="kotlin"
+                            src="snippets/tutorial-kotlin-rpc-server/src/main/kotlin/Main.kt"
+                            include-lines="18-20,26-49"
+                />
+                <p>
+                    Note that this implementation assumes the same <code>PizzaShopImpl</code> object will be shared
+                    across all clients, and ignores the thread safety issues this would cause. This will be addressed in
```
Mr3zee, line 557: Actually one `PizzaShopImpl` instance is created per one client instance. Still not thread-safe (at least for now), but the model is a bit different.

codeSnippets/snippets/tutorial-kotlin-rpc-client/src/main/kotlin/Main.kt
```diff
@@ -0,0 +1,60 @@
...
+    pizzaShop.orderPizza("CD34", Pizza("Margherita"))
+    pizzaShop.orderPizza("CD34", Pizza("Sicilian"))
+    pizzaShop.orderPizza("CD34", Pizza("California"))
+
+    streamScoped {
+        pizzaShop.viewOrders("AB12").collect {
+            println("AB12 ordered ${it.name}")
+        }
+
+        pizzaShop.viewOrders("AB12").collect {
```
Mr3zee, line 54: `viewOrders("CD34")`

Mr3zee (changes requested):
Hi, @vnikolova ! Thank you for this amazing tutorial!
I wrote some comments, though not show-stoppers, but still I think it would be great to address.
Please, let me know what you think in the comments!

meilalina:
Hi Viktoriya! Thank you very much for preparing this tutorial. A few comments, which I think can improve the experience and help us get conversions. 

1. Do you think we can remind to download IntelliJ IDEA at the beginning of the "Hello, kotlinx.rpc" section for those who immediately skip there after the intro? As an example it can be formulated as in this tutorial: https://kotlinlang.org/docs/jvm-get-started.html 

2. For now, I suggest we remove references to future articles: "Future articles will drill into low-level details and real-world considerations." and "Future articles will show how it would work with KMP-based desktop and mobile clients." as we don't know how soon and who can create them. 

3. In the current structure, I think that RPC vs REST and Kotlin RPC vs Google RPC chapters can be moved to the end (probably before the Next Steps chapter, though, WDYT?) as they are more general, and Extend the example still feels like the continuation of the tutorial. 

4. Let's provide options where people can share their feedback at the very end of the article: "so we encourage you to explore it and share your feedback with us." We can mention that bugs and feature requests can be submitted on YouTrack while general discussions take place on Slack. Here is the link to the channel for those who are already on Slack https://kotlinlang.slack.com/archives/C072YJ3Q91V , and the link for newcomers to get their invite: https://surveys.jetbrains.com/s3/kotlin-slack-sign-up 

Thank you!

codeSnippets/snippets/tutorial-kotlin-rpc-client/src/main/kotlin/Main.kt
vnikolova: The commented out parts are from the first iteration and are referenced in the tutorial. I wanted to avoid hard-coding them into the `.topic` file. Do you think it's better to create a separate file / code sample for each iteration?

topics/tutorial-first-steps-with-kotlin-rpc.topic
vnikolova: Reworded this a bit. Let me know if this is enough or if there's a link with more information we can include.

topics/tutorial-first-steps-with-kotlin-rpc.topic
vnikolova: Can you please explain what would be the correct procedure if we were to work with a single project from the beginning?
I do agree with you 100% that we should show the proper way of structuring. 
What if we keep this initial approach and add another iteration before the "Extend the example" section where we combine the modules under one project and create a shared one? In this way the reader can really feel the added value of single-sourcing.

vnikolova:
Thanks for the feedback @meilalina!

Regarding 1, since we have a 'Prerequisites' section, I don't think it is necessary to add this in the 'Hello, kotlinx.rpc' intro, but I added the same download link to the first step:

> 1. Launch [IntelliJ IDEA](https://www.jetbrains.com/idea/download/).

Another question arises though if we should provide an alternative approach to generating the projects as intelliJ IDEA is only a recommendation or should we mention the community edition as a minimum requirement?

2,3 and 4 - done ✅

topics/tutorial-first-steps-with-kotlin-rpc.topic
vnikolova: So is this accurate:
One `PizzaShopImpl` instance is created for each client instance. This avoids conflicts between clients but does not address thread safety within a single client's instance.

topics/tutorial-first-steps-with-kotlin-rpc.topic
Mr3zee: Yeah, that sound good. I would just replace `build time` with `compile time`, wdyt?

topics/tutorial-first-steps-with-kotlin-rpc.topic
Mr3zee: I think the absolutely best approach would be to have a Ktor Generator sample.

But since there is not one now, I think we should explain, that in one project you can have more than 1 module, exactly for such purposes. And we can even write `now create 3 identical folders: server, client and shared - and place this sample build.gradle.kts in all of them while all still in the same project`, but in more professional style. And the code snippet for these files would not be exactly big or complicated. And we should show the directories structure. Alternatively, same steps may be done from already exiting IDEA template singe-module project, where user just adds two more modules.

I strongly oppose the idea of creating 2 separate projects with copy-pasted services, as it is exactly what you wouldn't ever do with such setup. I agree, that is it way simpler to explain, but I consider it not that good in the long run

topics/tutorial-first-steps-with-kotlin-rpc.topic
Mr3zee: Yes. I just want to clarify the terminology a little. We have a client as `RPCClient`, which communicates with the `RPCServer`. Each client can have multiple services attached to it, even of the same type. This means that each of these services would have its own `Impl` instance on this server. So for two `PizzaShop` services where will be two different `PizzaShopImpl` instances on a server:
```kotlin
val service1 = client.withService<PizzaShop>()
val service2 = client.withService<PizzaShop>()

service1 != service2
```

So to sum up, saying `One PizzaShopImpl instance is created for each client instance` is not wrong, but `for each client service instance` would eliminate all ambiguity. 

This a technical part, which option is better to read and understand for a user is not obvious to me, so I would like to hear your opinion on it

topics/tutorial-first-steps-with-kotlin-rpc.topic
```diff
@@ -0,0 +1,599 @@
...
+                <path>Server.kt</path>
+                file and implement this functionality by storing the current orders in a map of lists:
+                <code-block lang="kotlin"
+                            src="snippets/tutorial-kotlin-rpc-app/src/main/kotlin/Server.kt"
+                            include-lines="18-20,26-49"
+                />
+                <p>
+                    Note that a new instance of <code>PizzaShopImpl</code> is created for each client service instance.
+                    This avoids conflicts between clients by isolating their state. However, it does not address thread safety within a single client's
+                    instance, particularly if the same instance is accessed concurrently by multiple coroutines.
```
Mr3zee, line 494: `it does not address thread safety within a single client's instance`

A small type, should be `within a single server's instance`, as the user state is present only on the server.

Other than that, looks great!

Mr3zee (approved):
Amazing, thank you!

PR #503, opened 2024-08-02
Update social preview image

zoobestik:
Like on other pages:
https://start.ktor.io/images/LinkPreviewImage.png


nikpachoo (approved):

vnikolova:
This doesn't look right. The image says "Ktor Generator", but the preview would be for the Ktor Documentation. Is there another image we could use?

PR #506, opened 2024-08-08
KTL-1547 Update social media preview images for Ktor web pages

nikpachoo:
https://youtrack.jetbrains.com/issue/KTL-1547


vnikolova (approved):
Looks good, thanks for the PR!

PR #509, opened 2024-08-11
Fix ref to code on client-create-and-configure.md

mister-proger:
https://ktor.io/docs/client-create-and-configure.html#plugins

The code in the example in this section of the documentation is non-working - there are no closing brackets. I have corrected the link to the lines so that the code is correct.


vnikolova (approved):
Thanks for the fix @mister-proger !

PR #511, opened 2024-08-14
Fix: KTOR-7248 Methods in FakeTaskRepository in first iteration shouldn't have the suspend keyword

Omasyo:
[KTOR-7248](https://youtrack.jetbrains.com/issue/KTOR-7248)

In this tutorial: https://ktor.io/docs/server-integrate-database.html#add-starter-code.

The TaskRepository interface does not match the sample code, it is missing the suspend keyword. I have fixed this.

### DIFF
--- a/topics/server-integrate-database.topic
+++ b/topics/server-integrate-database.topic
@@ -252,11 +252,11 @@
                         import Task
 
                         interface TaskRepository {
-                            suspend fun allTasks(): List<Task>
-                            suspend fun tasksByPriority(priority: Priority): List<Task>
-                            suspend fun taskByName(name: String): Task?
-                            suspend fun addTask(task: Task)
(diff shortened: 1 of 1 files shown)


vnikolova:
Thanks for picking this up @Omasyo. 

I had a look at the issue and it seems like there's been a mistake. It's not the `TaskRepository`, but the `FakeTaskRepository` in the next step that needs to be corrected. I updated the YT issue with the correct code snippet and updated description.

So, instead of referencing the code from the final example, we need to paste this part directly into the `<code-block>` element. Can you revert your changes and do this instead?

Omasyo:
Okay I'll do that

Omasyo:
How about this @vnikolova

vnikolova (approved):
Looks good, thanks for the update!

PR #516, opened 2024-08-23
Clarify Google AppEngine docs, support java21 runtime

kc9ddi:
The docs and code examples for deploying a Ktor server project to Google AppEngine currently demonstrates use of the `java11` run time environment. The java11 runtime will reach end of support within a few weeks (31 October 2024). (See the [Support Schedule](https://cloud.google.com/appengine/docs/standard/lifecycle/support-schedule#java) as published by Google.)

I've updated the example project to use the most recent support Java runtime (`java21`). This also requires use of the latest AppEngine plugin for Gradle, version 2.8.0. 

I've also made a few small revisions to the associated documentation to clarify a few items which I found confusing when I worked through this for my own project.

### DIFF
--- a/codeSnippets/snippets/google-appengine-standard/src/main/appengine/app.yaml
+++ b/codeSnippets/snippets/google-appengine-standard/src/main/appengine/app.yaml
@@ -1,2 +1,2 @@
-runtime: java11
+runtime: java21
 entrypoint: 'java -jar google-appengine-standard-1.0-SNAPSHOT-all.jar'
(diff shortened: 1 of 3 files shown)


codeSnippets/snippets/google-appengine-standard/src/main/appengine/app.yaml
```diff
@@ -1,2 +1,2 @@
-runtime: java11
+runtime: java21
 entrypoint: 'java -jar google-appengine-standard-1.0-SNAPSHOT-all.jar'
```
vnikolova, line 2: It seems that the entrypoint should also be updated to `java -jar google-appengine-standard-all.jar`

kc9ddi: You are correct, apologies for overlooking that. I've pushed a fix.

vnikolova (approved):
Looks good 👍  Thanks for the contribution @kc9ddi !

PR #517, opened 2024-08-23
Update README.md

lucasmsoares96:
update gradlew server argument

### DIFF
--- a/codeSnippets/snippets/tutorial-websockets-client/README.md
+++ b/codeSnippets/snippets/tutorial-websockets-client/README.md
@@ -8,7 +8,7 @@ A sample Ktor project showing how to create a [WebSocket chat client](https://kt
 Before running this sample, start a [Ktor server](../tutorial-websockets-server) first:
 
 ```bash
-./gradlew :tutorial-websockets-server:run
+./gradlew :tutorial-server-websockets:run
 ```
 
 Then, execute the following command to run a client sample:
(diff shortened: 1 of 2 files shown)


codeSnippets/snippets/tutorial-websockets-client/README.md
```diff
@@ -8,7 +8,7 @@ A sample Ktor project showing how to create a [WebSocket chat client](https://kt
 Before running this sample, start a [Ktor server](../tutorial-websockets-server) first:
```
vnikolova, line 8: Looks like the folder was renamed. Can you update the location here as well?

vnikolova (changes requested):
Thanks for the fix @lucasmsoares96! Just noticed one more place where the name needs to be updated.

codeSnippets/snippets/tutorial-websockets-client/README.md
lucasmsoares96: Should I change this line too?

https://github.com/ktorio/ktor-documentation/blob/9ec4bcabd2431d10fa1fb70d54b6745315c68e71/topics/client-create-websocket-chat.md?plain=1#L27

In this case, the link points to the file in version `2.3.4` in which the change was not applied. Should I change it to `main`?

vnikolova: Yes, well spotted! You can replace `2.3.4` in the URL with `%ktor_version%`. It will then automatically point to the relevant branch.

vnikolova (approved):
Looks good 👍

PR #518, opened 2024-08-27
Update server-integrate-database.topic

jaffar118406:
The path doesn't match default ktor project's path:
Changed from: com/example
Changed to: example/com


vnikolova:
Thanks for the contribution @jaffar118406! There were indeed inconsistencies in the paths, but it was the [Ktor Project Generator](https://start.ktor.io/settings) that was rendering incorrect folder structure.  This has now been fixed on the generator's end. `com/example` is the correct path.

PR #520, opened 2024-09-02
Fixed getting a non-existent parameter

yaroslavzghoba:
The form on the page `resources/static/index.html` designed to search for a task by name sends a GET request to the endpoint `/tasks/byName` with the name of the task in the `name` parameter. The endpoint, in turn, tries to get the value of the name by the `taskName` parameter. Now the GET method of the `/tasks/byName` endpoint tries to get the task name parameter by the name `name` instead of `taskName`.


vnikolova (approved):
Thanks for the fix @yaroslavzghoba !

PR #521, opened 2024-09-03
Delete the unused App1.kt

kez-lab:
Delete the unused App1.kt


kez-lab:
I realized that the App1.kt file is a guide. Thank you.

PR #523, opened 2024-09-05
Fix failing tests in tutorial samples

vnikolova:
The tests from the following samples were failing:

1. `tutorial-server-web-application`
The issue was caused by a missing `application.yaml` file.

2. `tutorial-server-db-integration`
Running the tests resulted in a`DuplicatePluginException` exception ([KTOR-7157](https://youtrack.jetbrains.com/issue/KTOR-7157)). To fix this I used the `MapApplicationConfig` function.
Additionally, updated the dependencies and imports to match the ones from the project generator, as well as the tutorial to include a final step for configuring the test application.

2. `tutorial-server-docker-compose`
This sample is using duplicated code from `tutorial-server-db-integration`. Because of this, the tests are not necessary.


e5l (approved):

topics/server-integrate-database.topic
```diff
@@ -867,21 +868,35 @@
...
+                <path>ApplicationTest.kt</path>
+                file, use the <code>MapApplicationConfig()</code> function to override the default application
+                configuration and use a custom module:
+                <code-block
+                        src="snippets/tutorial-server-db-integration/src/test/kotlin/com/example/ApplicationTest.kt"
+                        lang="kotlin"/>
+            </step>
+            <step>
+                <p>
+                    As the <code>FakeTaskRepository</code> type is no longer needed in production code, you can move
```
Stexxe, line 886: Why was it even there if it's a fake implementation?

Stexxe (approved):

topics/server-integrate-database.topic
vnikolova: It's considered useful for testing the routes only, though I do think having `Fake` in the name can be misleading. Perhaps we can call it `TestTaskRepository` or `InMemoryTaskRepository`?

PR #524, opened 2024-09-05
Included the helper function in the WebSocket configuration snippet.

yaroslavzghoba:
The `DefaultWebSocketServerSession.sendAllTasks` helper function is used in the code of the `Application.configureSockets` function, but it was not included in the topic. Because of its absence on the user side, the program code cannot even be compiled.

Before: 
Screenshot from 2024-09-05 16-23-02

After:
Screenshot from 2024-09-05 16-23-23


vnikolova (approved):
Thanks for the fix @yaroslavzghoba 🙏

PR #526, opened 2024-09-11
chore: replace com.github.johnrengelman.shadow with com.gradleup.shadow

MikAoJk:
com.github.johnrengelman.shadow plugin has changed the core maintainer and has new coordinates: https://plugins.gradle.org/plugin/com.gradleup.shadow

for more info se: https://plugins.gradle.org/plugin/com.github.johnrengelman.shadow


vnikolova (approved):
Thanks for the update @MikAoJk 🙏

PR #536, opened 2024-10-15
Fix typo in 3.0.x migration guide

FelixHarvey:
I noticed a very small typo when reading the migration guide.

`ApplciationEngine` should be `ApplicationEngine`.


vnikolova (approved):
Thanks for the fix @FelixHarvey !

PR #537, opened 2024-10-16
Document staticZip support of automatic reloading

vnikolova:
[KTOR-7311](https://youtrack.jetbrains.com/issue/KTOR-7311) 

There are no API changes, so I just added a subsection in the **Serving static content** topic about automatic reloading in `staticZip()`.
Also included the missing API links to the main functions mentioned.


e5l (approved):
looks good

PR #548, opened 2024-11-25
Fix minor typo in server-events.md

Martmists-GH:


vnikolova (approved):
Thanks for the fix @Martmists-GH !

PR #551, opened 2024-12-04
KTOR-7602 Fix deprecated date format in gson sample

bsautner:
[KTOR-7602 Fix deprecated date format in gson sample](https://youtrack.jetbrains.com/issue/KTOR-7602)

- Fixed deprecated usage of date format in Gson ContentNegotiation per JavaDoc:

> Counterintuitively, despite this method taking only a 'date style' Gson will use a format which includes both date and time, with the 'time style' being the last value set by setDateFormat(int, int). Therefore prefer using setDateFormat(int, int) and explicitly provide the desired 'time style'.

- Added unit test and a test gradle dependency for ContentNegotiation so I could properly test the response objects without using hardcoded json strings
- Added an Adapter for DateTime  to avoid an `com.google.gson.JsonIOException: Failed making field 'java.time.LocalDate#year' accessible;` error.


codeSnippets/snippets/gson/src/test/kotlin/com/example/GsonAppTest.kt
```diff
@@ -55,4 +60,28 @@ class GsonAppTest {
...
+                    setPrettyPrinting()
+                }
+            }
+        }
+
+        val response = client.get("/v1")
+        val responseModel = response.body<Model>()
+        assertEquals(HttpStatusCode.OK, response.status)
+        assertEquals(model, responseModel)
+
```
vnikolova, line 84: The two blank lines here can be removed.

vnikolova (commented):
Thanks for the fix @bsautner 🫶 
Looks good! Could you also update the `README.md` with the expected output?

codeSnippets/snippets/gson/src/test/kotlin/com/example/GsonAppTest.kt
bsautner: done :+1:

vnikolova (approved):

vnikolova:
Merged 🎉

bsautner:
:boom:  :heart_hands:

PR #555, opened 2024-12-08
KTOR-7222 POST call in tutorial should return 201 Created status

stokado:
At first I wanted to address only the ticket issue.
But it turned out, the tutorial was a little bit outdated.

I have carefully read the content of the "How to create RESTful APIs in Kotlin with Ktor﻿" tutorial and tested each step.

Current version of the [Project wizzard](https://start.ktor.io/settings?name=ktor-sample&website=example.com&artifact=com.example.ktor-sample&kotlinVersion=2.1.0&ktorVersion=3.0.2&buildSystem=GRADLE_KTS&buildSystemArgs.version_catalog=true&engine=NETTY&configurationIn=YAML&addSampleCode=true&plugins=routing%252Ccontent-negotiation%252Cstatic-content) does not create a "plugins" package, so I changed the routes and packages in the article.

There was also one mistake in the package name.

And I added one additional step in "Create unit tests with Ktor client" section, to add required dependency.
[YouTrack ticket](https://youtrack.jetbrains.com/issue/KTOR-7222/POST-call-in-tutorial-should-return-201-Created-status)

### DIFF
--- a/codeSnippets/snippets/tutorial-server-restful-api/src/main/kotlin/com/example/Application.kt
+++ b/codeSnippets/snippets/tutorial-server-restful-api/src/main/kotlin/com/example/Application.kt
@@ -1,6 +1,5 @@
 package com.example
 
-import com.example.plugins.*
 import io.ktor.server.application.*
 
 fun main(args: Array<String>) {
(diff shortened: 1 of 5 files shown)


vnikolova (changes requested):
Thanks for the fix and updating the steps @stokado! 💟 
Your changes look good, but there's still an import of `com.example.plugins.*` in `Application.kt` which is causing an error when building the project. Can you please remove it?

vnikolova (approved):
Great, looks ready to merge 🎉 
Thanks for your contribution @stokado !

PR #560, opened 2024-12-13
Fix micrometer-metrics running instruction in README.md

anton-erofeev:
Minor fix related to micrometer-metrics `README.md` `Running a sample` instruction


vnikolova (approved):
Thanks for the fix @anton-erofeev !

PR #561, opened 2024-12-16
KTOR-7560 Add the `content` property renaming of `HttpResponse` to the migration guide

vnikolova:
[KTOR-7560](https://youtrack.jetbrains.com/issue/KTOR-7560) 
Describe renaming from `content` to `rawContent` of `HttpResponse` in the migration guide


osipxd (approved):
👍

PR #565, opened 2024-12-28
`client-websockets`: add `ktor-client-cio` to required dependencies

BBBmau:
After going through the docs I found that it was necessary to include:
 ```implementation ("io.ktor:ktor-client-cio:3.0.2")```. 
 
 Just adding `implementation("io.ktor:ktor-client-websockets:3.0.2")` was not enough to get the example working due to missing dependencies.

Let me know if anything else is needed, this is just my assumption of where the change needs to be made.

Documentation: https://ktor.io/docs/client-websockets.html#add_dependencies

### DIFF
--- a/topics/client-websockets.topic
+++ b/topics/client-websockets.topic
@@ -9,7 +9,7 @@
     <var name="artifact_name" value="ktor-client-websockets"/>
     <tldr>
         <p>
-            <b>Required dependencies</b>: <code>io.ktor:ktor-client-websockets</code>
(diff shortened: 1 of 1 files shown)


vnikolova:
Hi @BBBmau, thank you for your initiative in contributing to Ktor Docs!

To clarify, working with the Ktor Client always requires an engine dependency, as explained in the [Adding client dependencies](https://ktor.io/docs/client-dependencies.html) topic. In the context of the WebSockets topic, the 'required dependencies' section refers to what’s additionally needed to use WebSockets with an existing Ktor Client. We used the CIO engine in the code examples, but you’re free to use any other engine that suits your needs.

If you have any further questions or feedback, let us know 🙏

PR #566, opened 2024-12-29
Make the dependency platform-specific

yuk1ty:
I encountered an error to inform that the app failed to load `application.conf`, instead of `application.yaml` when I set `ktor-server-config-yaml` to `build.gradle.kts`. This means the library doesn't work or just failed to read `application.yaml`. After some investigations, the following issue told me how I could resolve this error:

- https://youtrack.jetbrains.com/issue/KTOR-4757/Yaml-configuration-v2.1.0-does-not-seem-to-work

I think the dependency should be platform-specific. Otherwise it comes to a result of failing to load configuration file.

By the way, I'm a newbie for ktor and Kotlin. Please tell me if I overlooked something.

### DIFF
--- a/topics/server-configuration-file.topic
+++ b/topics/server-configuration-file.topic
@@ -63,7 +63,7 @@
                 </p>
                 <note>
                     <p>
-                        To use a YAML configuration file, you need to add the <code>ktor-server-config-yaml</code>
+                        To use a YAML configuration file, you need to add the <code>ktor-server-config-yaml-jvm</code>
                         <a href="server-dependencies.topic">
                             dependency
                         </a>


vnikolova:
Thanks for your initiative in contributing to Ktor Docs @yuk1ty! 🙏 
  
The -jvm suffix is ​​required only when using a Maven build system to explicitly select JVM-specific artifacts as Maven cannot guess the platform. In contrast, Gradle can automatically resolve platform-specific suffixes. Therefore, using the suffix is not always required. This is documented in the [Adding Ktor dependencies](https://ktor.io/docs/server-dependencies.html#add-ktor-dependencies) topic. 

If you have any further feedback or suggestions, please feel free to [open a new ticket on YouTrack](https://youtrack.jetbrains.com/issues/KTOR?q=Subsystem:%20Docs%20).

PR #572, opened 2025-01-04
Made $KTOR_ENV environment variable optional

LelouBil:
The environment variable is optional in the .conf file, and the code uses getPropertyOrNull, which implies the property is optional


vnikolova (approved):
Thanks for the fix @LelouBil !

PR #573, opened 2025-01-06
Fix server-routing.md documentation about greedy Regex

NinoDLC:
`Regex`es in Kotlin are greedy by default, so it would match the last `hello` in the path.


vnikolova (approved):
Thanks for the correction @NinoDLC !

PR #577, opened 2025-01-25
Fix: indentation in server-requests-and-responses.topic

mintobit:


vnikolova (changes requested):
Thanks for the PR @mintobit !
It looks like the indentation is wrong from line 395 as well. Can you please move the `if` and `try` blocks back by 1 stop?

mintobit:
@vnikolova so funny, I my brain noticed one curly brace and missed the whole two blocks xD

Adjusted. Thanks

vnikolova (approved):

PR #578, opened 2025-01-26
docs(server-create-a-new-project.topic): Fix unclosed brackets syntax error in Routing.kt

Ryuheeeei:
It seems that`Routing.kt` of https://ktor.io/docs/server-create-a-new-project.html#configure-static-content is broken due to lack of `}`. Besides that, I fix the indent as well.


vnikolova (approved):
Thanks for the fix @Ryuheeeei !

PR #513, opened 2024-08-16
KTOR-5085 Remove note about JVM-only support for XML serialization

osipxd:
Documentation change to reflect changes from https://github.com/ktorio/ktor/pull/4225

PR #579, opened 2025-01-28
KTOR-5085 Update info about XML serialization support

osipxd:
Changes in addition to #513


vnikolova (approved):
Thanks for adding this @osipxd !

PR #583, opened 2025-01-31
KTOR-7601: Fix dependencies for Maven build system tutorial

Ryuheeeei:
This PR fix [KTOR-7601](https://youtrack.jetbrains.com/issue/KTOR-7601/Update-dependency-and-Ktor-version-in-tutorial-server-get-started-maven)

* Following [Adding Ktor dependencies](https://ktor.io/docs/server-dependencies.html#add-ktor-dependencies), use platform-specific artifacts (i.e. `ktor-server-netty-jvm` and `ktor-server-test-host-jvm`) e81e3125594dc12857072dbc637c6013bda82bd2
  * This is the same as the dependencies which I choose 'Build System' as 'Maven' in Ktor project create page https://start.ktor.io/settings
* ktor and kotlin versions are same as [gradle.properties](https://github.com/ktorio/ktor-documentation/blob/main/codeSnippets/gradle.properties) respectively cd4a883057f816ac0326c7bdbb8df2effb29aa96

After this change, successfully I could build and run the application following README.md instructions.

(This is my first time to tackle with YouTrack issue, so please tell me if I missed any procedure. For example, should I assign the issue to me in advance?)

### DIFF
--- a/codeSnippets/snippets/tutorial-server-get-started-maven/pom.xml
+++ b/codeSnippets/snippets/tutorial-server-get-started-maven/pom.xml
@@ -8,9 +8,9 @@
     <name>tutorial-server-get-started-maven</name>
     <description>tutorial-server-get-started-maven</description>
     <properties>
-        <ktor_version>3.0.0-beta-1</ktor_version>
+        <ktor_version>3.0.3</ktor_version>
         <kotlin.code.style>official</kotlin.code.style>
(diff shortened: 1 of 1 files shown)


vnikolova (approved):
Thanks for picking this up @Ryuheeeei !
It is not a requirement, but it is useful if you comment on the YouTrack issue so we can update its status. Otherwise, you did everything by the book 👌

PR #603, opened 2025-03-15
fix: KTOR-8246: Docker example includes unnecessary copy statements i…

Eveeifyeve:
[KTOR-8246](https://youtrack.jetbrains.com/issue/KTOR-8246/Docs-Docker-example-includes-unnecessary-copy-statements-in-the-Dockerfile) 

* Remove unnecessary copy statements (/usr/src/app)


vnikolova (approved):
Thanks for the fix @Eveeifyeve !

PR #611, opened 2025-04-09
Update buildprofiles.xml with the current year so it propagates into the footer

siarb:


siarb:
@vnikolova Hi!
When you have a moment, could you please take a look and merge it? :)
(No rush - not urgent at all)

vnikolova (approved):
Hi @siarb ! Thanks for the update 🙏

PR #617, opened 2025-05-13
Remove redundant repeated paragraph.

skaldebane:
Also seems to be slightly different? The top one only mentions compressing _outgoing content_, while the one I removed mentioned compressing responses _and_ decompressing requests.

The docs don't say much about decompression, so I assume the top one is the latest. Correct me if I'm wrong!


Skaldebane:
Oops, sorry for the bot response. Didn't realize this thing kicks off everywhere :p

vnikolova:
Hi @Skaldebane ! Thanks for your PR and for noticing the redundancy 🙏 

While it is true that decompressing is not well documented, it is correct that the plugin provides both compressing responses and decompressing requests. Can you please use this version instead?

In the meantime, I've created a [new ticket](https://youtrack.jetbrains.com/issue/KTOR-8527) to track the missing documentation.

skaldebane:
@vnikolova that's good to know! Sorry for the late response.

Switched to the second paragraph. Should I open another PR?

vnikolova:
Sorry, @skaldebane, I closed it by mistake 😬 
Looks like the branch changed and I cannot reopen the PR, so please create a new one 🙏

skaldebane:
Done at #623!

PR #619, opened 2025-05-17
Update used snippet line in server uploads documentation

gpopides:
Looking at the files upload documentation, we can see that the included line in the example uses `PartData.FormItem` 
image

Which does not have for example a `.provider()` method.

I believe the line we want to include is line 25, which is the `is PartData.FileItem` part of the snippet

### DIFF
--- a/topics/server-requests.md
+++ b/topics/server-requests.md
@@ -161,7 +161,7 @@ In this example, the new limit is set to 100MB.
 ```kotlin
 ```
 
-{src="snippets/upload-file/src/main/kotlin/uploadfile/UploadFile.kt" include-lines="21,26-30,33"}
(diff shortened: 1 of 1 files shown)


vnikolova:
Hi @gpopides, thanks for noticing! 🙏 
You were on the right track, but the lines were outdated in multiple places, which was fixed with [this PR](https://github.com/ktorio/ktor-documentation/pull/620) so I'm closing yours. 

Btw, if you'd like to find more ways to contribute, feel free to check the [available tickets on YouTrack](https://youtrack.jetbrains.com/issues?u=1&q=project:%20Ktor%20subsystem:%20Docs%20tag:%20%7BUp%20For%20Grabs%7D) :)

PR #621, opened 2025-05-23
Fix typo in static content path example

4mjeo:
### Summary

This pull request fixes a small typo in the static content path example.  
The current documentation refers to `index.hml`, which seems to be a typo.  
It has been corrected to `index.html`.

### Why

Improving documentation accuracy and clarity.

---

Let me know if anything else needs to be updated.


vnikolova:
Thanks for the fix @4mjeo 🙏

vnikolova (approved):

PR #625, opened 2025-05-26
chore(docs): fix example path parameter

Liumingxun:
previous example cannot match "paramater is absent" case


vnikolova (approved):
Thanks for the update @Liumingxun, LGTM!

PR #629, opened 2025-06-03
Correcting sentence about missing implementation

hussilarabi:
The sentence "The only operation you are missing is the Delete." is not true, as Update was also not implemented till that point - only Read and Create.

### DIFF
--- a/topics/server-create-restful-apis.topic
+++ b/topics/server-create-restful-apis.topic
@@ -616,7 +616,7 @@
     <chapter title="Add support for removals" id="remove-tasks">
         <p>
             You have almost finished adding the basic operations to your service. These are often summarized as the CRUD
-            operations - short for Create, Read, Update, and Delete. Now You are going to implement the Delete operation.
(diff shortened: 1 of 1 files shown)


topics/server-create-restful-apis.topic
```diff
@@ -616,7 +616,7 @@
     <chapter title="Add support for removals" id="remove-tasks">
         <p>
             You have almost finished adding the basic operations to your service. These are often summarized as the CRUD
-            operations - short for Create, Read, Update, and Delete. The only operation you are missing is the Delete.
+            operations - short for Create, Read, Update, and Delete. Now You are going to implement the Delete operation.
```
vnikolova:
```suggestion
            operations - short for Create, Read, Update, and Delete. Now you are going to implement the Delete operation.
```

vnikolova (approved):
Thanks for the fix @hussilarabi ! 
Just fixed a small typo ('you' capitalization) bit it looks good otherwise! 👍

PR #630, opened 2025-06-03
New Configuration Pages

nomisRev:
This PR is a draft for new configuration pages. There is a lot of duplication, which seems to send you in a circle if you follow references. This PR attempts to simplify the configuration pages, and aims to add new practical information for building Ktor applications. 

1. Addressing the FAQ of _which engine to use_:
  - By explicitly highlight Netty with `tip`
  - Moving dependency configuration up, and moving different options down
2. Explain how to use Ktor Configuration to setup and customise users own configuration requirements
3. Add sections on how to deal with environment / system properties, and layered environment-specific configuration
4. Make more use of tabs to show different options side-by-side


topics/server-configuration.md
```diff
@@ -0,0 +1,244 @@
+[//]: # (title: Application Configuration)
+
+<show-structure for="chapter" depth="4"/>
+
+<link-summary>
+Learn how to configure various engine, and custom application parameters in code and file-based configurations.
+</link-summary>
+
+When running a Ktor server application, you might need to configure Ktor and application-specific configuration.
+These kinds of configurations are unrelated to [Engine Configuration][server-engines.md] and allow configuring Ktor
```
vnikolova, line 10: What do you mean by "configuring Ktor itself"? Is it different than server configuration?

ktor.tree
```diff
@@ -47,10 +47,9 @@
                 <toc-element topic="server-engines.md"
                              toc-title="Engines"
                              accepts-web-file-names="custom-engines.html,engines.html"/>
-                <toc-element topic="server-configuration-code.topic"
-                             accepts-web-file-names="configuration-code.html"/>
-                <toc-element topic="server-configuration-file.topic"
-                             accepts-web-file-names="configuration.html,configurations.html,environments.html,configuration-file.html"/>
+                <toc-element topic="server-configuration.md"
+                             toc-title="Server Configuration"
```
vnikolova, line 51: It's confusing that the TOC title is "Server Configuration" but the topic one "Application Configuration".  If I understand correctly, they are separate concepts, so we should either call this "Configuration" or "Server and application configuration". Wdyt?

topics/server-configuration.md
```diff
@@ -0,0 +1,244 @@
...
+itself or your own configurations required for your application.
+
+### Overview {id="configuration-overview"}
+
+Ktor supports loading [HOCON](https://github.com/lightbend/config/blob/master/HOCON.md) and [YAML](https://yaml.org)
+configuration files out-of-the-box.
+If you use [EngineMain](server-create-and-configure.topic#engine-main) to start a server, Ktor will automatically load
```
vnikolova, line 17: I think this deserves a sub-section, such as "Supported formats", now that the content is merged in one topic. 
We also previously mentioned that the files need to be located in the **resources** directory. Is this no longer a requirement?

topics/server-configuration.md
nomisRev: There is "Engine" configuration which is specific to the engine used "Netty", "CIO", etc.
Then there is Ktor specific configuration such as "ktor.development", "ktor.deployment.watch", "ktor.application.id", etc. Some overlap such as `port` and `host`.

These are currently _split_ from the framework point of view, and I kept them on `server-configuration` because that's where they lived. I should re-add `Predefined properties﻿` at the bottom of the page though.

ktor.tree
nomisRev: This is probably oversight. I thought `Server and application configuration` was too long for the menu 😅
I think `Server configuration`, `Application configuration`, or `Server and Application Configuration` all work.

`Server configuration` implies both "ktor server" and "my custom server logic", similarly for `Application` it refers to my own `Application` and the Ktor `Application` class.

I personally have no strong opinion. It could also be moved "Ktor Configuration" to "Engines" but then we'd have to change the name and I think that might be move confusing.

topics/server-configuration.md
nomisRev: > We also previously mentioned that the files need to be located in the resources directory.
Great point! This should still be included.

Good point about the sub-section!

ktor.tree
vnikolova: Alright then, I think it makes sense to keep the Ktor configurations here as opposed to the Engines topic. 
Maybe @meilalina can help with the title :)

PR #634, opened 2025-06-12
KTOR-8469 Add documentation about wrapWithContent and wrap deprecation

daniCsorbaJB:
[KTOR-8469](https://youtrack.jetbrains.com/issue/KTOR-8469) Documentation for HttpClientCall: Deprecate `wrapWithContent` and `wrap`

This update modifies:

- What's new in Ktor 3.2.0

### DIFF
--- a/topics/Whats-new-in-Ktor-3-2-0.md
+++ b/topics/Whats-new-in-Ktor-3-2-0.md
@@ -33,3 +33,30 @@ client.prepareGet("/some-file").execute { response ->
 </compare>
 
 This approach streams the response directly, preventing the body from being saved in memory.
+
+### The .wrapWithContent() and .wrap() extension functions are deprecated
+
+In Ktor 3.2.0, the [`.wrapWithContent()`](https://api.ktor.io/ktor-client/ktor-client-core/io.ktor.client.plugins.observer/wrap-with-content.html) and [`.wrap()`](https://api.ktor.io/ktor-client/ktor-client-core/io.ktor.client.plugins.observer/wrap.html) extension functions
+are deprecated in favor of the new `.replaceResponse()` function.
+
+The `.wrapWithContent()` and `.wrap()` functions replace the original response body with a `ByteReadChannel` that can only be read once.
+If the same channel instance is passed directly instead of a function that returns a new one, reading the body multiple times fails.
+This can break compatibility with [`SaveBodyPlugin`](https://api.ktor.io/ktor-client/ktor-client-core/io.ktor.client.plugins/-save-body-plugin.html?query=val%20SaveBodyPlugin:%20ClientPlugin%3CSaveBodyPluginConfig%3E),
+which relies on reading the response multiple times:
+
+```kotlin
+// Replaces the body with a channel decoded once from rawContent
+val decodedResponse = decode(response.rawContent)
+call.wrapWithContent(decodedResponse)
+// This fails on subsequent accesses
+```
+
+To avoid this issue, use the `.replaceResponse()` function instead.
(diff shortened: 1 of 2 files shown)


topics/Whats-new-in-Ktor-3-2-0.md
```diff
@@ -33,3 +33,30 @@ client.prepareGet("/some-file").execute { response ->
...
 This approach streams the response directly, preventing the body from being saved in memory.
+
+### The .wrapWithContent() and .wrap() extension functions are deprecated
+
+In Ktor 3.2.0, the [`.wrapWithContent()`](https://api.ktor.io/ktor-client/ktor-client-core/io.ktor.client.plugins.observer/wrap-with-content.html) and [`.wrap()`](https://api.ktor.io/ktor-client/ktor-client-core/io.ktor.client.plugins.observer/wrap.html) extension functions
+are deprecated in favor of the new `.replaceResponse()` function.
+
+The `.wrapWithContent()` and `.wrap()` functions replace the original response body with a `ByteReadChannel` that can only be read once.
+If the same channel instance is passed directly instead of a function that returns a new one, reading the body multiple times fails.
+This can break compatibility with [`SaveBodyPlugin`](https://api.ktor.io/ktor-client/ktor-client-core/io.ktor.client.plugins/-save-body-plugin.html?query=val%20SaveBodyPlugin:%20ClientPlugin%3CSaveBodyPluginConfig%3E),
```
osipxd: We've deprecated `SaveBodyPlugin` in this version and replaced it with an internal plugin. I think we can write something like: "this could break compatibility between different plugins reading the response body as the first plugin would consume the body".

topics/Whats-new-in-Ktor-3-2-0.md
```diff
@@ -33,3 +33,30 @@ client.prepareGet("/some-file").execute { response ->
...
+which relies on reading the response multiple times:
+
+```kotlin
+// Replaces the body with a channel decoded once from rawContent
+val decodedResponse = decode(response.rawContent)
+call.wrapWithContent(decodedResponse)
+// This fails on subsequent accesses
```
osipxd: In most cases it will return an empty content rather than fail
```kotlin
// Replaces the body with a channel decoded once from rawContent
val decodedBody = decode(response.rawContent)
val decodedResponse = call.wrapWithContent(decodedBody).response

decodedResponse.bodyAsText() // The first call returns the actual body
decodedResponse.bodyAsText() // Subsequent calls return an empty string
```

topics/Whats-new-in-Ktor-3-2-0.md
daniCsorbaJB: nice 👍 thank you for the info, I like it — would this slightly different version work as well:

This can break compatibility between different plugins accessing the response body, because the first plugin to read it consumes the body:

osipxd: Yes, looks good to me!

osipxd (approved):
LGTM!

topics/Whats-new-in-Ktor-3-2-0.md
```diff
@@ -33,3 +33,34 @@ client.prepareGet("/some-file").execute { response ->
 </compare>
 
 This approach streams the response directly, preventing the body from being saved in memory.
+
+### The .wrapWithContent() and .wrap() extension functions are deprecated
```
vnikolova:
```suggestion
### The `.wrapWithContent()` and `.wrap()` extension functions are deprecated
```

vnikolova (approved):
Looks great! Thanks for picking this one up 🙏

PR #635, opened 2025-06-13
What's new in Ktor 3.2.0: network and Gradle

ElviraMustafina:
- [KTOR-8542](https://youtrack.jetbrains.com/issue/KTOR-8542) Documentation for Support accessing resolved IP address on an instance of `io.ktor.network.sockets.InetSocketAddress`
- [KTOR-8211](https://youtrack.jetbrains.com/issue/KTOR-8211) Documentation for Support enabling development mode from the CLI

### DIFF
--- a/topics/Whats-new-in-Ktor-3-2-0.md
+++ b/topics/Whats-new-in-Ktor-3-2-0.md
@@ -2,7 +2,6 @@
 
 # What's new in Ktor 3.2.0
 
-<<<<<<< KTOR-8542-access-resolved-ip-address
 Here are the highlights for this feature release:
 
 * Version Catalog
@@ -10,8 +9,6 @@ Here are the highlights for this feature release:
 * First-class HTMX support
 * Suspendable module methods
 
-=======
->>>>>>> 3.2.0
 ## Ktor Client
 
 ### `SaveBodyPlugin` and `HttpRequestBuilder.skipSavingBody()` are deprecated
@@ -74,11 +71,10 @@ call.replaceResponse {
     decode(response.rawContent)
 }
 ```
-<<<<<<< KTOR-8542-access-resolved-ip-address
 
 ### Access resolved IP address
 
-You can now use the new `resolveAddress()` function on `io.ktor.network.sockets.InetSocketAddress` instances.
+You can now use the new `.resolveAddress()` function on `io.ktor.network.sockets.InetSocketAddress` instances.
 This function allows you to obtain the raw resolved IP address of the associated host:
 
 ```kotlin
@@ -89,7 +85,7 @@ val rawAddress = address.resolveAddress()
 It returns the resolved IP address as a `ByteArray`, or `null` if the address cannot be resolved.
 The size of the returned `ByteArray` depends on the IP version: it will contain 4 bytes for IPv4 addresses and 
 16 bytes for IPv6 addresses.
-On JS and Wasm platforms, `resolveAddress()` will always return `null`.
+On JS and Wasm platforms, `.resolveAddress()` will always return `null`.
 
 ## Gradle plugin
 
@@ -120,5 +116,3 @@ directly using a Gradle CLI flag:
 ```bash
 ./gradlew run -Pio.ktor.development=true
 ```
-=======
->>>>>>> 3.2.0
--- a/topics/Whats-new-in-Ktor-3-2-0.md
+++ b/topics/Whats-new-in-Ktor-3-2-0.md
@@ -2,6 +2,13 @@
 
 # What's new in Ktor 3.2.0
 
+Here are the highlights for this feature release:
+
+* Version Catalog
+* Dependency Injection
+* First-class HTMX support
+* Suspendable module methods
+
 ## Ktor Client
 
 ### `SaveBodyPlugin` and `HttpRequestBuilder.skipSavingBody()` are deprecated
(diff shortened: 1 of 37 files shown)


topics/Whats-new-in-Ktor-3-2-0.md
```diff
@@ -33,3 +40,46 @@ client.prepareGet("/some-file").execute { response ->
...
+
+You can now use the new `resolveAddress()` function on `io.ktor.network.sockets.InetSocketAddress` instances.
+This function allows you to obtain the raw resolved IP address of the associated host:
+
+```kotlin
+val address = InetSocketAddress("sample-proxy-server", 1080)
+val rawAddress = address.resolveAddress()
+```
+
+It returns the resolved IP address as a `ByteArray`, or `null` if the address cannot be resolved.
```
osipxd, line 85: Should we mention that it is 4 bytes for IPv4 and 16 bytes for IPv6?

topics/server-development-mode.topic
```diff
@@ -77,26 +77,12 @@
                 </li>
                 <li>
                     <p>
-                        If you run your application using a <a href="server-dependencies.topic">Gradle</a> task, you can pass
-                        <code>io.ktor.development</code> to the
-                        <code>applicationDefaultJvmArgs</code> property in a <path>build.gradle(.kts)</path> file:
+                        If you run your application using a <a href="server-dependencies.topic">Gradle</a> task,
+                        you can use a Gradle CLI flag:
```
osipxd: It is not the direct equivalent to the configuration previously described here. User will have to specify the flag every time he runs the application. The direct equivalent would be:
```kotlin
ktor {
    development = true
}
```
Our Gradle plugin will set the system property under the hood.
But it also worth mentioning that a user can pass a project property or a system property `io.ktor.development` when they run a gradle task. It is useful for one-time development mode enabling, so this bullet is also good to have.

topics/Whats-new-in-Ktor-3-2-0.md
```diff
@@ -64,3 +71,46 @@ call.replaceResponse {
...
+On JS and Wasm platforms, `resolveAddress()` will always return `null`.
+
+## Gradle plugin
+
+### Enabling development mode via the command line
```
osipxd:
```suggestion
### Enabling development mode
```
I think we can omit "command line" here as this section also  describes the way to enable the dev mode from the Gradle plugin.

osipxd (changes requested):
I'm sorry for the review delay 🙏

topics/server-development-mode.topic
ElviraMustafina: As far as I understand, the current structure is still valid:

- First, we show how to enable development mode with
   ```
   ktor {
       development = true
   }
   ```
- Second, how to enable development mode specifically when running your application.

Summoning @vnikolova as a file owner  🙏

vnikolova: @ElviraMustafina Yes exactly. Since the configuration file method is already mentioned in an earlier chapter, it is correct to have the previous example removed.

Maybe for added clarity we can mention the CLI property is for one-time use ,eg:
"If you run your application using a <a href="server-dependencies.topic">Gradle</a> task,you can enable development mode for a single run by passing a Gradle project property:"

topics/Whats-new-in-Ktor-3-2-0.md
```diff
@@ -64,3 +74,51 @@ call.replaceResponse {
     decode(response.rawContent)
 }
 ```
+<<<<<<< KTOR-8542-access-resolved-ip-address
+
+### Access resolved IP address
+
+You can now use the new `resolveAddress()` function on `io.ktor.network.sockets.InetSocketAddress` instances.
```
vnikolova:
```suggestion
You can now use the new `.resolveAddress()` function on `io.ktor.network.sockets.InetSocketAddress` instances.
```

topics/Whats-new-in-Ktor-3-2-0.md
```diff
@@ -64,3 +74,51 @@ call.replaceResponse {
...
+
+```kotlin
+val address = InetSocketAddress("sample-proxy-server", 1080)
+val rawAddress = address.resolveAddress()
+```
+
+It returns the resolved IP address as a `ByteArray`, or `null` if the address cannot be resolved.
+The size of the returned `ByteArray` depends on the IP version: it will contain 4 bytes for IPv4 addresses and 
+16 bytes for IPv6 addresses.
+On JS and Wasm platforms, `resolveAddress()` will always return `null`.
```
vnikolova:
```suggestion
On JS and Wasm platforms, `.resolveAddress()` will always return `null`.
```

topics/Whats-new-in-Ktor-3-2-0.md
```diff
@@ -2,6 +2,16 @@
 
 # What's new in Ktor 3.2.0
 
+<<<<<<< KTOR-8542-access-resolved-ip-address
```
vnikolova: Looks like the file didn't merge correctly 🤔

topics/server-development-mode.topic
osipxd: I'm sorry if I'm missing something, but I see only a section describing how to enable development mode in `application.conf`. While the snippet is the same it is two different ways to enable development mode: via `application.conf` and via `build.gradle.kts` script.

vnikolova: Oh thanks for clarifying @osipxd , looks like we missed the point 😄  @ElviraMustafina 👆

ElviraMustafina: Oops! I included two options for running with a Gradle task: using the `build.gradle.kts` file and the CLI flag

vnikolova (approved):

PR #643, opened 2025-06-20
Add documentation for DI

vnikolova:
- Add documentation for DI in the _What's new in Ktor 3.2.0_ topic
- Create a new _Dependency Injection_ plugin topic

### DIFF
--- a/topics/server-dependency-injection.md
+++ b/topics/server-dependency-injection.md
@@ -32,8 +32,8 @@ dependencies {
 
     // Function references
     provide<GreetingService>(::GreetingServiceImpl)
-    provide<BankService>(::BankServiceImpl)
-    provide(::BankTeller)
+    provide(BankServiceImpl::class)
+    provide(::createBankTeller)
 
     // Registering a lambda as a dependency
     provide<() -> GreetingService> { { GreetingServiceImpl() } }
@@ -66,7 +66,9 @@ class UserRepository(val db: Database) {
 }
 ```
 
-Arguments are resolved automatically through annotations like `@Property` and `@Named`.
+Ktor resolves constructor and function parameters automatically using the DI container. You can use annotations like
+`@Property` or `@Named` to override or explicitly bind parameters in special cases, such as when the type alone is not
+enough to distinguish a value. If omitted, Ktor will attempt to resolve parameters by type using the DI container.
 
 ## Dependency resolution and injection
 
@@ -88,7 +90,7 @@ To support asynchronous loading, you can use suspending functions:
 
 ```kotlin
 suspend fun Application.installEvents() {
-  val kubernetesConnection = dependencies.resolve() // suspends until provided
+  val kubernetesConnection: EventsConnection = dependencies.resolve() // suspends until provided
 }
 
 suspend fun Application.loadEventsConnection() {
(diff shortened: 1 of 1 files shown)


topics/server-dependency-injection.md
```diff
@@ -0,0 +1,255 @@
...
+
+```kotlin
+dependencies {
+    // Lambda-based
+    provide<GreetingService> { GreetingServiceImpl() }
+
+    // Function references
+    provide<GreetingService>(::GreetingServiceImpl)
+    provide<BankService>(::BankServiceImpl)
+    provide(::BankTeller)
```
bjhham: Maybe you could have:

```kotlin
provide<GreetingService>(::GreetingServiceImpl)
provide(BankServiceImpl::class)
provide(::createBankTeller)
```

This way we have an example of using a class reference and a normal function reference (non-constructor).

topics/server-dependency-injection.md
```diff
@@ -0,0 +1,255 @@
...
+// Repositories.kt
+fun provideDatabase(@Property("database.connectionUrl") connectionUrl: String): Database =
+  PostgresDatabase(connectionUrl)
+
+class UserRepository(val db: Database) {
+  // implementation 
+}
+```
+
+Arguments are resolved automatically through annotations like `@Property` and `@Named`.
```
bjhham: It's important to note here that the annotations are not required - they're only for special cases.  When absent, we resolve normally through the DI container.

topics/server-dependency-injection.md
```diff
@@ -0,0 +1,255 @@
...
+val service = dependencies.resolve<GreetingService>()
+```
+
+### Asynchronous dependency resolution
+
+To support asynchronous loading, you can use suspending functions:
+
+```kotlin
+suspend fun Application.installEvents() {
+  val kubernetesConnection = dependencies.resolve() // suspends until provided
```
bjhham: Needs a type here for compilation

```suggestion
  val kubernetesConnection: EventsConnection = dependencies.resolve() // suspends until provided
```

topics/server-dependency-injection.md
```diff
@@ -0,0 +1,255 @@
...
+```yaml
+ktor:
+  application:
+    dependencies:
+      - com.example.PrintStreamProviderKt
+    modules:
+      - com.example.LoggingKt.logging
+```
+
+```kotlin
```
bjhham, line 128: Probably want to include the dependencies function reference here for clarity.

Something like:
```
// com.example.PrintStreamProvider.kt
fun stdout(): () -> PrintStream = { System.out }
```

topics/server-dependency-injection.md
```diff
@@ -0,0 +1,255 @@
...
+
+```yaml
+connection:
+  domain: api.example.com
+  path: /v1
+  protocol: https
+```
+
+```kotlin
+val connection: Connection by application.property("connection")
```
bjhham: This one is just a regular function call

```suggestion
val connection: Connection = application.property("connection")
```

topics/server-dependency-injection.md
```diff
@@ -0,0 +1,255 @@
...
+```kotlin
+dependencies {
+  provide<List<String>> { listOf("one", "two") }
+}
+
+// This will work due to type parameter covariance support
+val stringList: List<CharSequence> by dependencies
+// This will also work
+val stringCollection: Collection<CharSequence> by dependencies
+```
```
bjhham: Might be good to expand this section to mention that covariance works with non-parameterized types as well, and that we don't support parameter sub-types for cases like resolving `Sink<CharSequence>` from `Sink<String>` even though they're technically covariant.

vnikolova: Added another example and 'Limitations' section.

PR #644, opened 2025-06-20
update: mention the Unix domain socket functionality in making requests

zamulla:


topics/client-default-request.md
```diff
@@ -99,6 +100,26 @@ defaultRequest {
...
+You can [build individual requests with Unix domain sockets](client-requests.md#specify-a-unix-domain-socket),
+but you can also configure a default request with a socket parameter.
+
+To do that, pass a `unixSocket` call with the path to the socket to the `defaultRequest` function,
+for example:
+
+```kotlin
+val response = HttpClient(CIO) {
+    defaultRequest {
+        unixSocket("/tmp/test-unix-socket-ktor.sock")
```
e5l, line 126: I would keep show `get("/") { unixSocket(...) } ` as a first sample instead of default request

topics/client-requests.md
```diff
@@ -119,6 +120,38 @@ You can configure a URL fragment using the `fragment` property.
...
+        }
+    }
+}
+```
+
+Then you can make requests to a server listening to that socket by calling the `unixSocket` function
+for a CIO client:
+
+```kotlin
+val response: HttpResponse = HttpClient(CIO).request {
```
e5l: HttpClient has its own lifecycle, so it should be closed. Please consider avoiding creating an HttpClient and executing the request right away.

I would also use some specific HTTP method (like `post/get` instead of a generic purpose request). The generic purpose request is not readable and is usually used for cases when the method is specified programmatically.

e5l (changes requested):
Hey @zamulla, thank you for the contribution!
Please check the comments

topics/client-default-request.md
zamulla: Do you mean show an individual request before getting into the `defaultRequest` example? I tried doing this, although it looks a bit redundant.

e5l: Yes, the `defaultRequest` is complexity we don't need from the beginning

e5l (approved):

topics/client-requests.md
```diff
@@ -119,6 +120,40 @@ You can configure a URL fragment using the `fragment` property.
 Note that `fragment` [encodes][percent_encoding] a URL fragment.
 To disable encoding, use `encodedFragment`.
 
+## Specify a Unix domain socket
+
+> Unix domain sockets are supported only in the CIO engine, for Ktor server as well as Ktor client.
```
vnikolova: This topic is regarding the Ktor Client only. The server-related documentation should be placed in _server-configuration-code.topic_

topics/client-default-request.md
```diff
@@ -99,6 +100,36 @@ defaultRequest {
 }
 ```
 
+### Unix domain sockets
+
+> Unix domain sockets are supported only in the CIO engine, for Ktor server as well as Ktor client.
```
vnikolova: This is a client plugin, so there's no need to mention Ktor server.

```suggestion
> Unix domain sockets are supported only in the CIO engine.
```

topics/client-requests.md
zamulla: I agree with the general point, but the server configuration doc doesn't list any options at all, only pointing at the API reference. So I thought it would be strange to manually describe a particular one now — tell me what you think about that logic.

Another point is that even if I did put a reference to Unix sockets in the server doc, I would still leave the client as is so that the user doesn't have to look at two pages to put the scenario together, which seemed an aggravating hypothetical to me.

vnikolova: But in this case you are assuming that the user is using a Ktor server as well, which might not be the case. I think it's ok to link to the relevant server section but not mix them in together. For the time being, you can just remove the server info from the client docs as the server configuration topic is being restructured anyway (as per #630).

zamulla: Yep, I am assuming that, you're right. I would like to at least link the server docs so that people have less trouble finding the relevant reference, but I'll remove the example.

topics/whats-new-320.md
```diff
@@ -203,6 +203,37 @@ Ktor’s HTMX support is available across three experimental modules:
 All APIs are marked with `@ExperimentalKtorApi` and require opt-in via `@OptIn(ExperimentalKtorApi::class)`.
 For more information, see [](htmx-integration.md).
 
+## Unix domain sockets
+
+With 3.2.0, you can set up Ktor clients to connect to Unix domain sockets and Ktor servers to listen to such sockets.
+Right now, Unix domain sockets are only supported in the CIO engine.
```
vnikolova:
```suggestion
Currently, Unix domain sockets are only supported in the CIO engine.
```

vnikolova (approved):
Looks good 👌

PR #645, opened 2025-06-21
fix(snippet): Fix OAuth Google snippet

jibidus:
The OAuth google snippet requires these modifications to work.


e5l:
Hey @jibidus, thanks for the PR. lgtm

PR #662, opened 2025-08-02
fix: Update the WebSocket configuration example code to use kotlin.time instead of java.time

juseong-park:
## Motivation
The current WebSocket plugin configuration example uses a `java.time.Duration` for `pingPeriod`, `timeout`, which is outdated due to recent interface changes. (https://github.com/ktorio/ktor/pull/4315)
The `pingPeriod`, `timeout` parameter now expects a `kotlin.time.Duration` type. To align with idiomatic Kotlin usage, it is better to use the `15.seconds` extension from the `kotlin.time` instead of calling `Duration.ofSeconds(15)`.

## Changes
Updated the WebSocket configuration example to use `15.seconds` for the `pingPeriod`, `timeout` setting.

### DIFF
--- a/topics/server-create-websocket-application.topic
+++ b/topics/server-create-websocket-application.topic
@@ -164,8 +164,8 @@
                         fun Application.configureSockets() {
                             install(WebSockets) {
                                 contentConverter = KotlinxWebsocketSerializationConverter(Json)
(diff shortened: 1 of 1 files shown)


vnikolova (approved):
Looks good! Thanks for the fix @juseong-park 🙏

PR #665, opened 2025-08-08
Mention linuxArm64 support

lukewarlow:
This updates the Ktor docs to add the missing mentions of linuxArm64, which is actually supported but was never documented.


vnikolova (approved):
Thanks for the addition @lukewarlow 🙏 
Looks good!

PR #666, opened 2025-08-10
fix: Update tutorial-client-kmm sample for Kotlin 2.1.20 Compose compatibility

tokizuoh:
### Background
This version (1.5.4) of the Compose Compiler requires Kotlin version 1.9.20 but you appear to be using Kotlin version 2.1.20 which is not known to be compatible.

### Solution

- Updates compose-compiler version to 2.1.20 (matches Kotlin version)
- Adds `kotlinCompose` plugin per [official migration guide](https://kotlinlang.org/docs/compose-compiler-migration-guide.html)

Since Kotlin 2.0+, Compose Compiler versions align with Kotlin versions and require the new plugin.

### Impact
tutorial-client-kmm sample only

### DIFF
--- a/codeSnippets/snippets/tutorial-client-kmm/gradle/libs.versions.toml
+++ b/codeSnippets/snippets/tutorial-client-kmm/gradle/libs.versions.toml
@@ -4,7 +4,6 @@ kotlin = "2.1.20"
 coroutines = "1.9.0"
 ktor = "3.2.3"
 compose = "1.6.8"
-compose-compiler = "2.1.20"
 compose-material3 = "1.2.1"
 androidx-activityCompose = "1.8.2"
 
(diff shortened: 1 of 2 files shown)


codeSnippets/snippets/tutorial-client-kmm/gradle/libs.versions.toml
```diff
@@ -4,7 +4,7 @@ kotlin = "2.1.20"
 coroutines = "1.9.0"
 ktor = "3.2.3"
 compose = "1.6.8"
-compose-compiler = "1.5.4"
+compose-compiler = "2.1.20"
```
vnikolova: You should be able to safely remove the version definition and its usage in `androidApp/build.gradle.kts`:

```kotlin
  composeOptions {
        kotlinCompilerExtensionVersion = libs.versions.compose.compiler.get()
    }
```

vnikolova (approved):
Thanks for the fix @tokizuoh! Looks good 👍

PR #669, opened 2025-08-21
Update CIO support for Js and WasmJs

vnikolova:
**Description**:
Update CIO support to include Js and WasmJs in Ktor Client and Server engines documentation.
Additionally, improve structure and formatting in client-engines.md.

**Related tickets**:

[KTOR-7050](https://youtrack.jetbrains.com/issue/KTOR-7050) Documentation for Ktor client for Kotlin/Wasm
[KTOR-7728](https://youtrack.jetbrains.com/issue/KTOR-7728) Documentation for Client CIO engine support for wasm-js and js
[KTOR-8047](https://youtrack.jetbrains.com/issue/KTOR-8047) Documentation for Support CIO server on WasmJS and JS targets

### DIFF
--- a/topics/client-engines.md
+++ b/topics/client-engines.md
@@ -6,32 +6,34 @@
 Learn about engines that process network requests.
 </link-summary>
 
-The [Ktor HTTP client](client-create-and-configure.md) can be used on different platforms, including JVM, [Android](https://kotlinlang.org/docs/android-overview.html), [JavaScript](https://kotlinlang.org/docs/js-overview.html), and [Native](https://kotlinlang.org/docs/native-overview.html). A specific platform may require a specific engine that processes network requests. 
-For example, you can use `Apache` or `Jetty`for JVM applications, `OkHttp` or `Android` for Android, `Curl` for desktop applications targeting Kotlin/Native, and so on. Different engines may have specific features and provide different configuration options.
+The [Ktor HTTP client](client-create-and-configure.md) is multiplatform and runs on JVM,
+[Android](https://kotlinlang.org/docs/android-overview.html), [JavaScript](https://kotlinlang.org/docs/js-overview.html)
+(including WebAssembly), and [Native](https://kotlinlang.org/docs/native-overview.html) targets. Each platform requires
+a specific engine to process network requests.
+For example, you can use `Apache` or `Jetty` for JVM applications, `OkHttp` or `Android`
+for Android, `Curl` for desktop applications targeting Kotlin/Native. Every engine differs slightly in features and
+configuration, so you can choose the one that best meets your platform and use-case needs.
 
+## Supported platforms {id="platforms"}
 
-## Requirements and limitations {id="requirements"}
+The table below lists the [platforms](client-supported-platforms.md) supported by each engine:
 
-### Supported platforms {id="platforms"}
-
-The table below lists [platforms](client-supported-platforms.md) supported by each engine:
-
-| Engine  | Platforms                                                                                     |
-|---------|-----------------------------------------------------------------------------------------------|
-| Apache  | [JVM](#jvm)                                                                                   |
-| Java    | [JVM](#jvm)                                                                                   |
-| Jetty   | [JVM](#jvm)                                                                                   |
-| Android | [JVM](#jvm), [Android](#jvm-android)                                                          |
-| OkHttp  | [JVM](#jvm), [Android](#jvm-android)                                                          |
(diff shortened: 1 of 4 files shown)


topics/client-engines.md
```diff
@@ -60,181 +67,236 @@ The table below shows whether a specific engine supports HTTP/2 and [WebSockets]
...
 ```
-{src="snippets/_misc_client/DefaultEngineCreate.kt"}
 
-This can be useful for multiplatform projects. For example, for a project targeting both [Android and iOS](client-create-multiplatform-application.md), you can add the [Android](#jvm-android) dependency to the `androidMain` source set and the [Darwin](#darwin) dependency to the `iosMain` source set. The necessary dependency will be selected at compile time.
+{src="snippets/_misc_client/DefaultEngineCreate.kt"}
 
+This is especially useful in multiplatform projects. For example, for a project targeting
+both [Android and iOS](client-create-multiplatform-application.md), you can add the [Android](#jvm-android) dependency
+to the `androidMain` source set and the [Darwin](#darwin) dependency to the `iosMain` source set. The appropriate
+engine is selected at compile time.
```
osipxd: Oh, this wasn't true. The engine is selected at run time, at the moment of `HttpClient` creation

topics/client-engines.md
```diff
@@ -60,181 +67,236 @@ The table below shows whether a specific engine supports HTTP/2 and [WebSockets]
...
 ### Apache {id="apache"}
-The `Apache` engine supports HTTP/1.1 and provides multiple configuration options. 
-You can also use the `Apache5` engine if you need HTTP/2 support, which has HTTP/2 enabled by default.
 
-1. Add the `ktor-client-apache5` or `ktor-client-apache` dependency:
+The `Apache` engine supports HTTP/1.1 and provides multiple configuration options.
+If you need HTTP/2 support, use the [`Apache5` engine](#apache5), which enables it by default.
 
-   <var name="artifact_name" value="ktor-client-apache5"/>
+1. Add the `ktor-client-apache` dependency:
```
osipxd: I think we can skip the `Apache` engine from documentation and show only `Apache5`. We're going to deprecate `ktor-client-apache` ([KTOR-6766](https://youtrack.jetbrains.com/issue/KTOR-6766))

topics/client-engines.md
```diff
@@ -274,29 +338,28 @@ To use the `WinHttp` engine, follow the steps below:
...
-   ```bash
-   sudo apt-get install libcurl4-gnutls-dev
-   ```
-
-   > On Windows, you may want to consider the [MinGW/MSYS2](FAQ.topic#native-curl) `curl` binary. 
+    - On Linux, install the gnutls version of libcurl:
+    ```bash
+    sudo apt-get install libcurl4-gnutls-dev
+    ```
+    - On Windows, you may want to consider the [MinGW/MSYS2](FAQ.topic#native-curl) `curl` binary.
```
osipxd: This step is not required for Ktor >= 3.1.0

topics/client-engines.md
```diff
@@ -355,45 +427,59 @@ The `Js` engine can be used for [JavaScript projects](https://kotlinlang.org/doc
...
-         }
-      }
+   config(this)
+
+   engine {
+    config {
+        retryOnConnectionFailure(true)
+        connectTimeout(0, TimeUnit.SECONDS)
+        }
+     }
```
osipxd: Is indentation here correct?

osipxd (commented):
Nicely don!
Let's fix problems mentioned in the comments as well

topics/client-engines.md
vnikolova: Is `Apache` not supported by the client anymore? And does `Apache5` also support HTTP/1.1?

osipxd: The `Apache5` engine is a newer version of the `Apache` engine. The `Apache` engine depends on Apache HttpClient 4 which is deprecated. So we keep `ktor-client-apache` only for backwards compatibility and new users should use `ktor-client-apache5`

> does Apache5 also support HTTP/1.1?

Yes, it does support HTTP/1.1 and HTTP/2

topics/client-engines.md
vnikolova: No, you're right it's not. Should be fixed now.

topics/client-engines.md
vnikolova: Thanks for the confirmation. I've removed the `Apache` engine from the list, but added a note about the deprecation.

PR #673, opened 2025-09-09
Update `server-session-auth.md` to reference correct code snippet for configuring sessions

thijsnulle:
The current code snippet in `topics/server-session-auth.md` is:

```
cookie<UserSession>("user_session") {
        cookie.path = "/"
        cookie.maxAgeInSeconds = 60
    }
}
```

which corresponds correctly to lines 17-21 of `Application.kt`, however, it should also include the `install(Sessions) {` of line 16. For the code snippet see [here](https://ktor.io/docs/server-session-auth.html#install-session).

### DIFF
--- a/topics/server-session-auth.md
+++ b/topics/server-session-auth.md
@@ -81,7 +81,7 @@ installs and configures a cookie session with a specified cookie path and expira
 ```kotlin
 ```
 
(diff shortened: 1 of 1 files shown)


vnikolova (approved):
Thanks for the fix @thijsnulle! Looks good 👍

PR #674, opened 2025-09-09
KTOR-8822 Documentation for SSE response body buffer

vnikolova:
[KTOR-8822](https://youtrack.jetbrains.com/issue/KTOR-8822) 

- Add a new section in the "What's new in Ktor 3.3.0" topic.
- Add a new sub-section in the Configuration chapter of the Client's SSE plugin page.


topics/client-server-sent-events.topic
```diff
@@ -78,6 +78,60 @@
...
+                        <title><code>LastEvents(n)</code></title>
+                        Keeps the last n completed SSE events.
+                    </def>
+                    <def id="buffer-all">
+                        <title><code>All</code></title>
+                        Keeps all processed events so far.
+                    </def>
+                </deflist>
+                <p>
+                    On failure, the buffered snapshot is wrapped into a <code>SavedHttpResponse</code>. This allows safe
```
marychatte: Can we please not mention this part, because `SavedHttpResponse` is an internal class? We can just say that this buffer will be accessible with  `response?.bodyAsText()`

topics/client-server-sent-events.topic
```diff
@@ -78,6 +78,60 @@
             <code-block lang="kotlin" src="snippets/client-sse/src/main/kotlin/com.example/Application.kt"
                         include-lines="20-23"/>
         </chapter>
+        <chapter title="Response buffering" id="response-buffering">
+            <p>
+                SSE responses are streaming by nature, which makes capturing the full body impractical. You can enable
+                a diagnostic buffer to safely retrieve the response body when an SSE stream fails.
```
marychatte: Maybe we can mention here that the buffer represents already processed data (no re-reading from the network). It is intended for logging and error analysis when failures happen.

topics/client-server-sent-events.topic
```diff
@@ -78,6 +78,60 @@
...
+                    <def id="buffer-lastevent">
+                        <title><code>LastEvent</code></title>
+                        Keeps the last completed SSE event.
+                    </def>
+                    <def id="buffer-lastevents">
+                        <title><code>LastEvents(n)</code></title>
+                        Keeps the last n completed SSE events.
+                    </def>
+                    <def id="buffer-all">
+                        <title><code>All</code></title>
```
marychatte, line 127: Can we add a note for `All` option: Use with care for long-lived streams.

PR #676, opened 2025-09-09
KTOR-8845 Document limitations of auto-reload

vnikolova:
[KTOR-8845](https://youtrack.jetbrains.com/issue/KTOR-8845)

- add note about regression and chapter with more info in "What's new in 3.2.0"
- confirm the current state and future plans in "What's new in 3.3.0"
- add a _Limitations_ section to the auto-reload topic

### DIFF
--- a/topics/server-auto-reload.topic
+++ b/topics/server-auto-reload.topic
@@ -40,7 +40,58 @@
             </p>
         </li>
     </list>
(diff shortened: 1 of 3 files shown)


bjhham (approved):
Thanks for the quick update 🎉

PR #681, opened 2025-09-17
KTOR-8862 HTTP/2 without TLS

ElviraMustafina:
[KTOR-8862](https://youtrack.jetbrains.com/issue/KTOR-8862)

- Add a new section to the What's new in Ktor 3.3.0 page.
- Add a new sub-section in the Netty﻿ chapter of the HTTP/2 page.

### DIFF
--- a/codeSnippets/snippets/http2-netty/src/main/kotlin/com/example/Application.kt
+++ b/codeSnippets/snippets/http2-netty/src/main/kotlin/com/example/Application.kt
@@ -5,7 +5,14 @@ import io.ktor.server.request.*
 import io.ktor.server.response.*
 import io.ktor.server.routing.*
 
-fun Application.main() {
+fun main() {
+    embeddedServer(Netty, port = 8080, configure = {
+        enableHttp2 = true
+        enableH2c = true
+    }, module = Application::module).start(wait = true)
+}
+
+fun Application.module() {
     routing {
         get("/") {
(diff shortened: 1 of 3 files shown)


codeSnippets/snippets/http2-netty/src/main/kotlin/com/example/Application.kt
```diff
@@ -5,7 +5,14 @@ import io.ktor.server.request.*
 import io.ktor.server.response.*
 import io.ktor.server.routing.*
 
-fun Application.main() {
+fun main() {
+    embeddedServer(Netty, port = 8080, configure = {
```
vnikolova: This example now produces errors. Looks like [`embeddedServer`](https://api.ktor.io/ktor-server/ktor-server-core/io.ktor.server.engine/embedded-server.html) doesn't accept `port` and the `configure` param together.
Also new imports are missing and to use `embeddedServer` here, we also need to update the `application` config in **build.gradle.kts**.
I couldn't get it to work locally. @bjhham can you help?

bjhham: Good point!

```kotlin
embeddedServer(Netty, configure = {
    connector {
        port = 8080
    }
    enableHttp2 = true
    enableH2c = true
})
```

vnikolova (approved):
Looks good 👍

PR #682, opened 2025-09-17
KTOR-8867 Add documentation for WebRTC client

vnikolova:
[KTOR-8867](https://youtrack.jetbrains.com/issue/KTOR-8867)

- Add an entry in _What's new in Ktor 3.3.0_.
- Create a new topic _WebRTC client_ under _Setting up a client_.

### DIFF
--- a/topics/client-webrtc.md
+++ b/topics/client-webrtc.md
@@ -32,18 +32,19 @@ build features such as:
 
 ## Add dependencies {id="add-dependencies"}
 
-To use `WebRTClient`, you need to include the `%artifact_name%` artifact in the build script:
+To use `WebRtclient`, you need to include the `%artifact_name%` artifact in the build script:
 
 <include from="lib.topic" element-id="add_ktor_artifact"/>
 
 ## Create a client
 
-When creating a `WebRtcClient`, you need to choose an engine depending on your platform:
+When creating a `WebRtcClient`, choose an engine based on your target platform:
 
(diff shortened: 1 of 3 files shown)


topics/client-webrtc.md
```diff
@@ -0,0 +1,224 @@
...
+
+In addition to data channels, WebRTC supports media tracks for audio and video. This allows you to build applications
+such as video calls or screen sharing.
+
+### Creating local tracks
+
+You can request audio or video tracks from local devices (microphone, camera):
+
+```kotlin
+val audio = jsClient.createAudioTrack { echoCancellation = true }
```
zibet27: Sorry, I've used this method in the KLIP document, but I've just found that it's broken. The problem is that fields in the `WebRtcMedia.AudioTrackConstraints`(the same for video) are `val`, so `jsClient.createAudioTrack { echoCancellation = true }` will not compile.
I will create a pull request with a fix today, but for now, I think we should mention only this version:

```
val audioConstraints = WebRtcMedia.AudioTrackConstraints(
      echoCancellation = true
)
val videoConstraints = WebRtcMedia.VideoTrackConstraints(
      width = 1280,
      height = 720
)
val audio = rtcClient.createAudioTrack(audioConstraints)
val video = rtcClient.createVideoTrack(videoConstraints)
```

vnikolova: Ah thanks for clarifying. Updated the example!

zibet27 (approved):

PR #796, opened 2026-04-21
KTOR-9496 Docs for JavScript ES module support

vnikolova:
Add a what's new entry for the following issues:

[KTOR-9350](https://youtrack.jetbrains.com/issue/KTOR-9350) JS: Make ES2015 the default target for tests
[KTOR-7659](https://youtrack.jetbrains.com/issue/KTOR-7659) Make ktor-network compatible with ES modules for nodejs

### DIFF
--- a/topics/whats-new-350.md
+++ b/topics/whats-new-350.md
@@ -107,3 +107,15 @@ val config = ApplicationConfig("application.yaml").getAs<Config>()
 ```
 
 </compare>
+
+### ES modules compatibility for `ktor-network`
+
+We’ve fixed issues that made it impossible to use `ktor-network` and all dependent modules when ES modules were enabled.
+
+To help prevent future regressions, our JavaScript test infrastructure now targets both ES2015 and ES modules by default.
+
+> For more information about Kotlin/JS module systems and ES2015 support, see:
+> * [JavaScript modules](https://kotlinlang.org/docs/js-modules.html)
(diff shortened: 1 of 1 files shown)


topics/whats-new-350.md
```diff
@@ -79,3 +79,17 @@ install(Authentication) {
...
+
+Ktor test infrastructure now targets the [ES2015](https://262.ecma-international.org/6.0/) standard to verify compatibility
+with modern JavaScript environments and upcoming changes in Kotlin’s default JavaScript target. This ensures compatibility
+with ES module–based setups and keeps Ktor aligned with the evolving Kotlin/JS ecosystem.
+
+#### Node.js ES modules support in `ktor-network`
+
+The `ktor-network` module now works correctly in Node.js projects that use ES modules. This improves compatibility with
+modern bundlers when targeting both Node.js and browser environments and avoids issues caused by Node.js-specific
+modules (such as `net`) being included in browser builds.
```
osipxd: I'm sorry for not providing a proper description for the issues. I think we can omit details and write something like this instead of two sub-sections:
> We've fixed problems making it impossible to use ktor-network and all dependent modules with ES modules enabled. We've also updated our test infrastructure to target ES2015 and ES modules to prevent possible regressions.

These links might be helpful:
- https://kotlinlang.org/docs/js-project-setup.html#support-for-es2015-features
- https://kotlinlang.org/docs/js-modules.html

vnikolova: Thanks, updated the section now.

osipxd (approved):
👍

PR #804, opened 2026-04-30
KTOR-9516 Add docs for reading configuration as data classes using .getAs()

vnikolova:
- Add a what's new entry
- Add a new sub-section in Server's configuration in a file topic

[KTOR-9516](https://youtrack.jetbrains.com/issue/KTOR-9516)

### DIFF
--- a/topics/server-configuration-file.topic
+++ b/topics/server-configuration-file.topic
@@ -477,10 +477,14 @@
                 }
             }
         ]]></code-block>
-        <chapter title="Read structured configuration" id="structured-config">
+        <chapter title="Deserialize configuration into data classes" id="deserialize-config">
             <p>
-                If your configuration is structured, you can deserialize it into Kotlin data classes using the
-                <code>.getAs()</code> function.
+                You can deserialize configuration into Kotlin data classes using the
+                <code>.getAs()</code> function. This approach provides type-safe access to configuration values.
+            </p>
+            <p>
+                The example below defines <code>app</code> and <code>security</code> configuration sections and maps
+                them to serializable Kotlin data classes.
             </p>
             <tabs group="config">
                 <tab title="application.conf" group-key="hocon" id="config-conf-1">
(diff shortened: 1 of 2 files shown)


topics/server-configuration-file.topic
```diff
@@ -480,9 +477,57 @@
                 }
             }
         ]]></code-block>
+        <chapter title="Read structured configuration" id="structured-config">
+            <p>
+                If your configuration is structured, you can deserialize it into Kotlin data classes using the
```
Stexxe: How can it not be structured?

topics/server-configuration-file.topic
```diff
@@ -480,9 +477,57 @@
...
+            @Serializable
+            data class Config(val app: App, val security: Security)
+
+            fun Application.module() {
+                val config = environment.config.getAs<Config>()
+
+                val port = config.app.port
+                val host = config.app.host
+
+                println("Starting server on $host:$port")
```
Stexxe: I get that it's for illustration purposes, but a similar message is already logged on the server startup.

topics/server-configuration-file.topic
```diff
@@ -480,9 +477,57 @@
...
+            }
+            ]]></code-block>
+        </chapter>
         <p>
-            This is especially useful when you keep <a anchor="custom-property">custom settings</a> in a configuration
-            file and need to access its values.
+            Reading configuration in code is especially useful when you keep <a anchor="custom-property">custom settings</a>
```
Stexxe: What are the other options if this one is especially useful?

vnikolova: Not reading it, I guess 😅 
Removed it as it's unnecessary.

topics/server-configuration-file.topic
vnikolova: Changed it to read the `clientId` and `clientSecret` instead. Let me know if you have better ideas for an example if this doesn't make sense.

topics/server-configuration-file.topic
vnikolova: True, it's not specific enough. Updated the title and the intro to be less ambiguous.

Stexxe (approved):

PR #813, opened 2026-05-14
KTOR-7600 Fix broken /free-api link in rate-limit README

fru1tworld:
[KTOR-7600](https://youtrack.jetbrains.com/issue/KTOR-7600) (item 2)

`Application.kt` defines `/public-api` (under `RateLimitName("public")`), not `/free-api`, so the URL listed in the README returns 404.

Items 1 and 3 from the ticket are not addressed here — left a note on YouTrack with my findings.


vnikolova (approved):
Thanks for the fix @fru1tworld !

PR #814, opened 2026-05-14
KTOR-8228 Use yyyy instead of YYYY in logback date pattern

fru1tworld:
[KTOR-8228](https://youtrack.jetbrains.com/issue/KTOR-8228)

The logback configs included by the [server-logging](https://ktor.io/docs/server-logging.html) docs page used `YYYY-MM-dd` (ISO week-based year). Switch to `yyyy-MM-dd` (calendar year), which is what the issue reporter requested.

Other snippets in the repo also use `YYYY` but are out of scope for this issue.


vnikolova (approved):
Thanks for the update @fru1tworld. 
Looks good 👍

PR #817, opened 2026-05-18
KTOR-9042 Document the required trailing slash on the default base path

fru1tworld:
[KTOR-9042](https://youtrack.jetbrains.com/issue/KTOR-9042)

Add a note to the Base URL section stating that the default base path must end with a trailing slash; otherwise the request path replaces the last segment per RFC 3986. Existing example and snippet already use a trailing slash, so no code changes are needed.


vnikolova (approved):
Thanks for the addition @fru1tworld !
Looks good 👍

PR #819, opened 2026-05-19
KTOR-9593 Update the Create a WebSocket application tutorial

vnikolova:
- Update file names and code blocks.
- Update snippets project.
- Update plugins screenshot.

[KTOR-9593](https://youtrack.jetbrains.com/issue/KTOR-9593)


codeSnippets/snippets/tutorial-server-websockets/src/main/kotlin/Routing.kt
```diff
@@ -1,16 +1,45 @@
...
+            sendAllTasks()
+            close(CloseReason(CloseReason.Codes.NORMAL, "All done"))
+        }
+
+        webSocket("/tasks2") {
+            sessions.add(this)
+            sendAllTasks()
+
+            while(true) {
+                val newTask = receiveDeserialized<Task>()
```
Stexxe, line 31: We might want to call `ensureActive()` at the beginning of the loop body to return early if the scope is closed.

codeSnippets/snippets/tutorial-server-websockets/src/main/kotlin/Routing.kt
```diff
@@ -1,16 +1,45 @@
...
         }
-        // Static plugin. Try to access `/static/index.html`
         staticResources("/static", "static")
     }
 }
+
+private suspend fun DefaultWebSocketServerSession.sendAllTasks() {
+    for (task in TaskRepository.allTasks()) {
+        sendSerialized(task)
+        delay(1000.milliseconds)
```
Stexxe, line 45: Why delay?

codeSnippets/snippets/tutorial-server-websockets/gradle.properties
```diff
@@ -0,0 +1,9 @@
+#Kotlin
+kotlin.code.style=official
+kotlin.daemon.jvmargs=-Xmx8g
+
+#Gradle
+org.gradle.jvmargs=-Xmx12g -Dfile.encoding=UTF-8
```
Stexxe: Why so much memory?

bjhham: Looks like this was introduced in the generator.  I'll put in a fix https://github.com/ktorio/ktor-plugin-registry/pull/262/

codeSnippets/snippets/tutorial-server-websockets/src/main/kotlin/Routing.kt
vnikolova: > For demonstration purposes, a one-second delay is introduced between sending tasks. This allows us to observe the tasks appearing incrementally in our client. Without this delay, the example would appear identical to the [RESTful service](https://ktor.io/docs/server-create-restful-apis.html) and [Web Application](https://ktor.io/docs/server-create-website.html) developed in previous articles.

bjhham (approved):

PR #827, opened 2026-06-09
KTOR-9538 Add docs for session ID generation  and clearing sessions by ID

vnikolova:
[KTOR-9538](https://youtrack.jetbrains.com/issue/KTOR-9538)


zibet27 (changes requested):
Nice!

CodeRabbit has a couple of useful comments

zibet27 (approved):

PR #835, opened 2026-07-08
Fix wording on property delegation in DI

bjhham:
The lazy evaluation behaviour was later changed to ensure that all declarations are checked during the startup phase.  Discovered this after [a comment from a user](https://youtrack.jetbrains.com/issue/KTOR-8329/Dependencies-should-be-validated-on-startup#focus=Comments-27-14006857.0-0).


vnikolova (approved):
Thanks for the update @bjhham 🙏

PR #836, opened 2026-07-09
KTOR-9494 Add docs for request decompression and clarify pipeline

vnikolova:
[KTOR-9494](https://youtrack.jetbrains.com/issue/KTOR-9494)

Add a new section for decompression in the `Compression` plugin page.

### DIFF
--- a/topics/server-compression.md
+++ b/topics/server-compression.md
@@ -45,8 +45,6 @@ To include Zstandard compression, add the `ktor-server-compression-zstd` depende
 <include from="lib.topic" element-id="install_plugin"/>
 
 This enables the `gzip`, `deflate`, and `identity` encoders on a server.
-In the next chapter, we'll see how to enable only specific encoders and configure conditions for compressing data.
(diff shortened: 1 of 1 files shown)


topics/server-compression.md
```diff
@@ -157,6 +155,22 @@ install(Compression) {
...
+to `CompressionConfig.Mode.DecompressRequest`:
+
+```kotlin
+install(Compression) { 
+    mode = CompressionConfig.Mode.DecompressRequest
+    gzip()
+}
+```
+
+When request decompression is enabled, the plugin uses the configured encoders to decompress the request body with the
```
Stexxe: 1. I know in the code they are called "encoders", but it's strange to read. It should be decoders instead.
2. The "with" part is a bit ambiguous because it could mean decompressing the request body with a tool, not that the request has the `Content-Encoding` header.

vnikolova: Good points - updated the sentence.

bjhham (approved):

PR #837, opened 2026-07-16
feat: KTOR-9518 Add a code example for the property() function

vnikolova:
[KTOR-9518](https://youtrack.jetbrains.com/issue/KTOR-9518)

### DIFF
--- a/topics/server-configuration-file.topic
+++ b/topics/server-configuration-file.topic
@@ -479,7 +479,7 @@
         ]]></code-block>
         <chapter title="Deserialize configuration into data classes" id="deserialize-config">
             <p>
-                You can deserialize configuration into Kotlin data classes for type-safe access to configuration values.
+                You can deserialize configuration into Kotlin classes for type-safe access to configuration values.
             </p>
             <p>
                 The example below defines <code>app</code> and <code>security</code> configuration sections and maps
--- a/topics/server-configuration-file.topic
+++ b/topics/server-configuration-file.topic
@@ -479,8 +479,7 @@
         ]]></code-block>
         <chapter title="Deserialize configuration into data classes" id="deserialize-config">
             <p>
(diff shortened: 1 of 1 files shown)


topics/server-configuration-file.topic
```diff
@@ -479,8 +479,7 @@
         ]]></code-block>
         <chapter title="Deserialize configuration into data classes" id="deserialize-config">
             <p>
-                You can deserialize configuration into Kotlin data classes using the
-                <code>.getAs()</code> function. This approach provides type-safe access to configuration values.
+                You can deserialize configuration into Kotlin data classes for type-safe access to configuration values.
```
zibet27: Afaik, regular classes are also supported

topics/server-configuration-file.topic
```diff
@@ -511,6 +510,23 @@
                 ]]></code-block>
                 </tab>
             </tabs>
+            <p>
+                Use the <code>Application.property()</code> function to deserialize a specific configuration section:
+            </p>
+            <code-block lang="kotlin"><![CDATA[
+            @Serializable
+            data class Security(val clientId: String, val clientSecret: String)
```
zibet27, line 518: Should we mention more complex and edge cases?

vnikolova: We can. Do you have any examples for such?

Stexxe (approved):

bjhham (approved):

topics/server-configuration-file.topic
zibet27: For instance, what happens if there is no such property (afair exception is thrown) or the property is of invalid shape (a deserialization exception is thrown; also, afaiu users can't configure `kotlinx.serialization` used to parse config and its format specific)
Maybe worth mentioning `Application.propertyOrNull` extension

PR #840, opened 2026-07-20
KTOR-9548: Update the log4j bridge version in Logging

stokado:
Updated slf4j bridge for log4j


vnikolova (approved):
Cool, thanks for the update 👍

PR #841, opened 2026-07-21
KTOR-7600 Fix tomcat-war-ssl sample so it actually runs

stokado:
tomcat-war-ssl snippet changes

### Related issue

[KTOR-7600](https://youtrack.jetbrains.com/issue/KTOR-7600) Update READMEs in ktor-documentation projects


zibet27 (approved):
Looks good to me

vnikolova (approved):
Thanks for the fix @stokado 👏

PR #853, opened 2026-08-24
KTOR-9795 Add docs for the ktor-client-engine-defaults artifact

vnikolova:
**Description:**
- Add a what's new entry.
- Edit the multiplatform engine section in _Adding dependencies_.
- Add a new default engines sub-section in _Client engines_.

**Related issue:**
[KTOR-9795](https://youtrack.jetbrains.com/issue/KTOR-9795) Documentation for Client curated multi-platform facade module

### DIFF
--- a/topics/whats-new-360.md
+++ b/topics/whats-new-360.md
@@ -69,6 +69,40 @@ It is now deprecated in favor of `.respondHtmlPartial()`.
 
 ## Ktor Client
 
+### Default client engines for multiplatform projects
+
+Ktor 3.6.0 introduces the `ktor-client-engine-defaults` artifact, which provides a curated set of HTTP [client engines](client-engines.md)
+for [Kotlin Multiplatform](https://kotlinlang.org/docs/multiplatform/get-started.html) projects.
+
+Add the dependency to the `commonMain` source set:
+
+```kotlin
+kotlin {
+    sourceSets {
+        commonMain {
+            dependencies {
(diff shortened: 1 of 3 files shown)


bjhham (approved):

topics/whats-new-360.md
```diff
@@ -69,6 +69,41 @@ It is now deprecated in favor of `.respondHtmlPartial()`.
...
+
+For each target platform, Ktor uses the default engine provided by `ktor-client-engine-defaults`. This removes the
+need to declare separate engine dependencies for each platform in most multiplatform projects.
+
+If your multiplatform project currently uses `CIO` across all supported targets, consider replacing the `CIO` dependency
+with `ktor-client-engine-defaults`. This lets Ktor provide a curated default engine for each platform while keeping
+engine selection out of your common source set.
+
+You can still [declare a specific client engine](client-dependencies.md#kmp-specific-engine) when you need engine-specific
+configuration or behavior.
```
bjhham, line 106: We've also updated the behaviour of the `HttpClient()` function so that when there is more than one engine available, it will construct a client from the engine with the highest priority.  By default, CIO has the lowest priority, so it will not be used when there's more than one engine available for the given platform.  Previously, the selection of engine was random and gave no feedback towards which was chosen.

vnikolova: Thanks for clarifying. I've mentioned the updated behavior in the docs.

PR #864, opened 2026-09-01
KTOR-9840 Add docs for nullable request bodies with call.receive()

vnikolova:
**Description**
- Add a what's new entry.
- Mention using nullable request bodies in the _Objects_ subsection of the _Server requests_ topic.
- Fix formatting, structure, and language in the entire _Server requests_ topic.

**Related issue**
[KTOR-9840](https://youtrack.jetbrains.com/issue/KTOR-9840) Documentation for Support nullable types in ApplicationCall.receive

### DIFF
--- a/topics/whats-new-360.md
+++ b/topics/whats-new-360.md
@@ -123,6 +123,58 @@ routing {
 You can also place `rateLimit()` outside `authenticate()` to apply rate limiting before authentication. Use this approach
 when the rate limit doesn't depend on an authenticated principal.
 
+### Nullable request bodies with `ApplicationCall.receive()`
+
+Ktor now supports nullable type arguments with the `ApplicationCall.receive()` function.
+
+The `.receiveNullable()` function is deprecated. Use `.receive()` with a nullable type when a request body can be `null`:
+
+<compare first-title="3.5.x" second-title="3.6.0">
+
+```kotlin
+ post("/") {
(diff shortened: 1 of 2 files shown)


topics/whats-new-360.md
```diff
@@ -123,6 +123,58 @@ routing {
...
+
+Ktor now supports nullable type arguments with the `ApplicationCall.receive()` function.
+
+The `.receiveNullable()` function is deprecated. Use `.receive()` with a nullable type when a request body can be `null`:
+
+<compare first-title="3.5.x" second-title="3.6.0">
+
+```kotlin
+ post("/") {
+    val payload = call.receiveNullable<Payload>()
```
osipxd: `receiveNullable` should've also been called with a nullable type. Otherwise its behavior matches `receive<T>()`. 
```suggestion
    val payload = call.receiveNullable<Payload?>()
```

So in other words `receive<T>()`, ` receiveNullable<T>()` both don't deserialize to `null`, and the only correct use of `receiveNullable` is `receiveNullable<T?>()`. That was the confusion we tried to solve with the updated API.

osipxd (approved):
Great! I have only one small comment

topics/whats-new-360.md
vnikolova: Ah got it, thanks for clarifying 🙏

PR #866, opened 2026-09-02
Update WebRTC Docs after version update.

zibet27:
We updated the version of webrtc-java, and now, there is no limitation with the `BufferedAmountLow` event


vnikolova (approved):
Thanks for the update!
Looks good 👌

